PR #28, opened 2018-11-14
feat: define an extensions section of the manifest

technosophos:
This defines a section of the manifest where arbitrary extensions can be added.

The goal of this section is to provide a specific location for extension data to be defined such that we don't end up like `package.json`, where tons of ad hoc extensions have been added to the root document.

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -453,7 +453,7 @@ Implementations that do not support custom actions MUST NOT emit errors (either 
 
 In many cases, the bundle descriptor is a sufficient artifact for delivering a CNAB bundle, since invocation images and other images may be retrieved from registries and repositories. Consequently, it is important to provide an extension mechanism. An _extension_ is a named collection of auxiliary data whose meaning is defined outside of this specification.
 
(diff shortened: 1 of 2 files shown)


michelleN:
-1 for metadata, info
+1 for ext

technosophos:
In the TUF spec, they use `custom`, which I sorta like as it gets right to the point.

simongdavies:
+1 for custom

garethr:
+1 for custom, although extensions also works for me

Note that this PR should also update the JSON Schema with the new fields.

101-bundle-json.md
```diff
@@ -453,5 +463,54 @@ The built-in actions (`install`, `upgrade`, `uninstall`) MUST NOT appear in the
...
+    }
+}
+```
+
+The format is:
+
+```json
+{
+    "extensions": {
+        "EXTENSION NAME": "ARBITRARY JSON DATA"
```
glyn: This gives the impression that `"ARBITRARY JSON DATA"` must be double quote delimited. Is the intention that this should be a JSON-encoded string or can it be arbitrary "inline" JSON?

glyn: I guess the examples resolve that question, but best to have some normative text too.

technosophos: If only JSON had comments...

radu-matei: No one say YAML. Please.

technosophos:
Since the original PR was before the repo renames, I had to open a new version of this PR at #104. So that should supersede this one. I am closing this one.

PR #46, opened 2018-11-30
Introducing namespacing to the names

### LINKED ISSUE #45: Namespacing names
The names need to be namespaced. Without namespacing we can run into collision issues and confusion.

I'm assuming we know why namespacing is needed. I'm happy to go into detail if anyone needs it.

Before one or more proposed implementations come in as a pull request I thought it would be good to capture some requirements in an issue that can be referenced.

1. Namespacing should be 2 or more levels deep. For example, `samsung.wordpress` and `samsung.sds.wordpress` are both valid because some organizations have nested orgs

2. The same name should be valid if the package is pushed to more than one repository. This essentially means we cannot use a URL as the endpoint. This is needed, for example, to handle signed packages in mirrors or for organizations with multiple registries

mattfarina:
Closes #45

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -6,9 +6,9 @@ A `bundle.json` is broken down into the following categories of information:
 
 - The schema version of the bundle, as a string with a `v` prefix. This schema is to be referenced as `v1` or `v1.0.0-WD`
 - The top-level package information (`name` and `version`)
-    - name: The bundle name
-    - version: Semantic version of the bundle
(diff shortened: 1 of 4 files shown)


technosophos:
@mattfarina can haz rebase?

mattfarina:
@technosophos something broke in this issue because of the flip of this repo from private to public after the original PR. The link between this and my fork was broken. I had to create a new PR at #99.

Issue #58, opened 2018-12-16
image substitution


itaysk:
Images used by the application are referenced in the bundle and substituted by the runtime. At first I thought about substitution scenarios which are not defined in the spec like case sensitivity, substrings, wildcards, and more.. But then it occurred to me that it's better to just use regexp substitution, possibly with the familiar `sed` format. For simple uses it won't impose too much burden on the user, but it will enable advanced uses with ease.

itaysk:
also addresses https://github.com/deislabs/cnab-spec/blame/master/101-bundle-json.md#L269

simonferquel:
For some scenarios, declarative substitutions might not be enough (e.g.: replacements inside heavily templated files that are not valid yaml/json/toml, or replacements in compiled code data structures). For this, I suggest that complementary to the declarative approach, we also inject a map of `component name -> image name` in a well known location of the invocation image at run time. PR https://github.com/deislabs/duffle/pull/605 on duffle implement the idea, and a correlated PR in docker-app leverages that as well.

technosophos:
I think this one is now resolved by #62

PR #62, opened 2018-12-18
Introduce image map injection

simonferquel:
This brings 2 changes:
- Bundle.json `images` field is now a map (`component logical name` -> `image definition`)
- `images` map is injected at runtime so that the invocation image can use it when declarative substitutions are not applicable

Relates to discussion in #58


technosophos:
@simonferquel if you are good with this as-is, I will merge it.

simonferquel:
@technosophos no need for the invocation image I think (do we have scenarios where the invocation image must instanciate itself ?)
We still need to decide if we want to keep the declarative reference manipulation or not.

technosophos:
Alright, I am merging this PR.

PR #73, opened 2019-01-23
Require Canonical JSON for bundle.json

### LINKED ISSUE #72: Use Canonical JSON
We should use Canonical JSON so that hashing is consistent across implementations.

technosophos:
Closes #72

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -2,6 +2,10 @@
 
 This section describes the format and function of the `bundle.json` document.
 
+The `bundle.json` file is a representation of bundle metadata. It MUST be represented as [Canonical JSON](http://wiki.laptop.org/go/Canonical_JSON). While Canonical JSON is parseable by any JSON parser, its serialized form is consistent. This is a necessity when comparing two textual representations of the same data (such as when hashing).
+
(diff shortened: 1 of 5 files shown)


101-bundle-json.md
```diff
@@ -2,6 +2,10 @@
 
 This section describes the format and function of the `bundle.json` document.
 
+The `bundle.json` file is a representation of bundle metadata. It MUST be represented as [Canonical JSON](http://wiki.laptop.org/go/Canonical_JSON). While Canonical JSON is parseable by any JSON parser, its serialized form is consistent. This is a necessity when comparing two textual representations of the same data (such as when hashing).
```
lachie83: Should we add appropriate verbiage around whether or not tooling should error if the bundle.json is not represented in the correct format?

lachie83:
What are the implications of merging this change? In my mind they are benign in the short-term although should tools error if bundle.json is non-compliant given this change?

101-bundle-json.md
technosophos: This part is tricky. I haven't seen any Canonical JSON parsers that actually verify that the content it parses is canonical. But we could certainly add a MAY line in here.

technosophos:
Yeah, so my thought is that since Canonical JSON parsers don't appear to error out if the JSON is not perfectly canonical (e.g. "non-validating parsers"), this change actually impacts generators first. So we can ease our way into it.

PR #75, opened 2019-01-24
Address various ambiguities in spec and schema

### LINKED ISSUE #66: Questions on specificity of the specifcation
Before spotting #36 I wrote up a sample JSON Schema for `bundle.json`, and jotted down a few ambiguities or questions I came across. See  https://gist.github.com/garethr/5e45b35bff950ad3b62dd98fe592f3d3 for the schema.

1. is `schemaVersion` required?
2. If `schemaVersion` is not required, what is the expected behaviour?
3. The example in the spec has a `uri` field for images, but the spec doesn't mention this field?
4. For field selectors section referring to other specs (like the rough [JsonPath](https://goessner.net/articles/JsonPath/), [JSON Pointer - RFC6901](https://tools.ietf.org/html/rfc6901) and XPath) might be wise
5. Should `invocationImages` be required?
6. Should `additionalProperties` be allowed that are not defined in the spec?
7. `string, int, boolean` for `parameters.type` shortens `int` but doesn't shorten `boolean`

Happy to submit PRs to address these once we have answers.

technosophos:

Closes #66

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -190,8 +190,9 @@ The following fields are informational pieces of metadata designed to convey add
 
 ## Invocation Images
 
-The `invocationImages` section describes the images that contains the bootstrapping for the image. The appropriate invocation
-image is selected using the current driver.
+The `invocationImages` section describes the images that contains the bootstrapping for the image. The appropriate invocation image is selected using the current driver.
+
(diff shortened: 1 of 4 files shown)


technosophos:
@garethr I can keep updating this per issue #66 until we have all the items suitably addressed.

PR #77, opened 2019-01-24
fix: switch to JSON Schema for parameter validation.

### LINKED ISSUE #54: Lean on JSON Schema for parameter validation?
The spec has had parameter validation since fd3d65a0ffe (#1), but the existing constraints are fairly limited and are maybe unique to this specification.  Can we use JSON Schema's [validation keywords][1] (or a subset of those keywords) instead of rolling a custom validation format?

[1]: https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-6

### LINKED ISSUE #55: Parameter validation
Parameters specify basic validation rules (e.g. min/max length, type). I'm wondering why not use JSON schema for a complete rule set which is standard, comprehensive and easy to implement?

technosophos:
This switches the validator keywords to a subset of the ones used in JSONSchema.

Closes #54
Closes #55

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -327,14 +327,17 @@ Parameter specifications are flat (not tree-like), consisting of name/value pair
 - parameters: name/value pairs describing a user-overridable parameter
   - `<name>`: The name of the parameter. This is REQUIRED. In the example above, this is `backend_port`. This
     is mapped to a value definition, which contains the following fields:
-    - type: one of string, int, boolean (REQUIRED)
+    - type: "null", "boolean", "string", "number", or "integer". These types correspond to the main primitive types of JSON (with _integer_ as a special case). Objects and arrays are not supported, as the serialization of these is underdetermined. (REQUIRED)
     - required: if this is set to true, a value MUST be specified (OPTIONAL, not shown)
     - defaultValue: The default value (OPTIONAL)
-    - allowedValues: an array of allowed values (OPTIONAL)
-    - minValue: Minimum value (for ints) (OPTIONAL)
-    - maxValue: Maximum value (for ints) (OPTIONAL)
+    - enum (alias allowedValues): an array of allowed values (OPTIONAL)
+    - minimum (alias minValue): Minimum value (for numeric and integer) (OPTIONAL)
+    - exclusiveMinimum: Exclusive minimum value (for numeric and integer) (OPTIONAL) 
+    - maximum (alias maxValue): Maximum value (for numeric and integer) (OPTIONAL)
+    - exclusiveMaximum: Exclusive maximum value (for numeric and integer) (OPTIONAL)
     - minLength: Minimum number of characters allowed in the field (for strings) (OPTIONAL)
     - maxLength: Maximum number of characters allowed in the field (for strings) (OPTIONAL)
+    - pattern: An ECMA 262 regular expression that must match the string (OPTIONAL)
     - metadata: Holds fields that are not used in validation (OPTIONAL)
       - description: A user-friendly description of the parameter
     - destination: Indicates where (in the invocation image) the parameter is to be written (REQUIRED)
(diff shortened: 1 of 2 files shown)


technosophos:
Note that this (if merged) will require changes to Duffle, Docker Apps, and Porter -- as well as any other ecosystem tools.

technosophos:
@itaysk Does this look better to you than our previous attempt?

itaysk:
yep, thanks!

Unrelated to this PR but related to coupling with JSON schema, solves this TODO I believe: https://github.com/deislabs/cnab-spec/blame/a51d5b11700b076625bedb9ac08dfa73481adfc6/104-claims.md#L98
`TODO: What is the best timestamp format to use? Does JSON have a preference?`

JSON does not but JSON schema does: https://json-schema.org/latest/json-schema-validation.html#rfc.section.7.3.1

technosophos:
Excellent! Thanks!

bacongobbler:
When we eventually drop the deprecated fields, does that entail a version bump to the spec to v2?

I'm all for this change; I'd just like to know how downstream clients handle CNABs that conform to the old spec when we make this change. I'm assuming this wouldbe the following migration plan:

For simplicity's sake, I'll henceforth refer to Duffle, Porter, Docker Apps, etc. "CNAB Clients".

- CNAB Clients generate CNABs with the new validator keywords
- CNAB Clients MUST accept both `minValues` (old CNABs) and `minimum` (new CNABs)
- when the old properties are eventually dropped (for example `minValues`), a new schema version bump is issued. Once a CNAB Client implements v2, they will have to drop support for "legacy" CNABs.

Would it be better to just drop those fields now and issue v2 of the CNAB spec? I'm not sure what's the right migration plan here, so I'd love to hear from others on what's the right approach here.

carolynvs:
Since it's a breaking change, yeah I hope we do bump to v2. For porter, we don't have anything to support and are riding the YOLO wave. So we will pretty quickly jump to v2 and only support that (no legacy spec support). 

cc @jeremyrickard in case he violently disagrees

garethr:
I don't think we should rev the spec version. It's a working draft, as per the process https://github.com/deislabs/cnab-spec/blob/master/901-process.md#process and we should be happy with making breaking changes.

The PR doesn't deprecate this, it changes it. We should focus on the spec over the tools in my view in the short term, otherwise we're paying a large cost for zero or minimal benefit in my view.

youreddy:
Hey @technosophos! Was taking a look at this PR and noticed [this line](https://github.com/deislabs/cnab-spec/pull/77/files#diff-0e61ada784c4ce6e546107b972180d42R298). I'm curious as to why serialization is undetermined for objects and arrays, could you elaborate?

/cc @ryanmoran @madamkiwi

technosophos:
Not _undetermined_, but _underdetermined_. That is, there is not enough information about how these things should behave to warrant supporting them as first-class objects. The preferred route is to handle embedding arbitrary JSON by serializing it into a string prior to sending it into the parameter field.

There are two related problems:

- JSON does not have a fixed field ordering for objects, which means that injecting raw JSON Into Canonical JSON will cause a problem. But assuming that the far-side implementation can handle data serialized with Canonical JSON is (slightly) presumptuous.
- More importantly, it is unclear what would be done with objects and arrays when injected into env vars or files.  Do we serialize them? How? What serialization scheme do we use to ensure compatibility with an unknown destination environment?
- Compounding the second one, i's unclear how we would apply constraints to arbitrary JSON data. All of the other types can be constrained according to fairly straightforward rubrics. But arrays and objects cannot. We'd end up having to support something like arbitrary JSON schema.

So rather than make the parameter specification highly complex, we just opted to not support them. If people desire to nest JSON objects inside of a parameter field, that can be done by serializing the JSON objects and injecting them as strings. Then, those people can choose what the serialization strategy is for themselves.

ryanmoran:
> So rather than make the parameter specification highly complex, we just opted to not support them. If people desire to nest JSON objects inside of a parameter field, that can be done by serializing the JSON objects and injecting them as strings. Then, those people can choose what the serialization strategy is for themselves.

Does this not just end in a spec within the spec and divergent parameter formats?

On another front, how do you imagine these parameter inputs scaling to hundreds or thousands of options?

I would imagine the burden of finding and loading the contents of hundreds of properties from different environment locations would be burdensome on a bundle author. At that scale, does the burden not rise to the same level as specifying that the inputs would arrive within the invocation container as JSON?

technosophos:
I don't understand. I just said that supplying JSON is fine -- just store it as an encoded string. What I am saying is that translating an on-disk representation to arbitrary JSON, then translating it for an unknown runtime requires domain knowledge that is non-generic. If you have thousands of parameters, I would STRONGLY suggest putting them in a config file of some sort, and passing the entire config.

More concisely, I'm arguing that if you have structured config data, CNAB shouldn't have to know anything about it other than that it needs to be passed from a source to a destination.

ryanmoran:
I understand that you are proposing that bundle authors could have users encode and pass JSON through a string configuration parameter. I think that this type of parameter input will likely be a common one. There are many examples I can think of where a finite set of strings, booleans, or integers cannot fulfill the configuration requirements of an application.

Additionally, from your further explanation, I think it is clear that the `string` type creates no real constraints upon the inputs. In fact, the inputs all serialize to strings at some point within the invocation image as environment variables and files have no real enforcement of type.

I guess the question I have then is, why does the spec make an effort to provide types and validation for some inputs (`string`, `int`, `boolean`) but not for others (`array`, `object`)? For example, why does the spec make a meaningful determination between the input of a port number (supported fully by the `int` type) versus an unbounded list of ports (requiring a bespoke serialization format)?

technosophos:
I'd be okay with that if there was an implementation that could show how the marshal and unmarshal is done consistently. If you'd like to do those as separate PRs -- one here and one on Duffle -- I'd be in favor of it. As I said in my first response today, I was just shooting for simplicity.

So I'm assuming that what you're going to propose is using JSON Schema for arbitrarily complex parameters? And then unmarshal input, validate it, and re-marshal it into the destination? I could see doing that -- as long as it doesn't make the bundle.json absurdly complex. Perhaps you might consider storing those schemata external to the bundle.json?

Anyway, happy to review any PRs.

PR #83, opened 2019-01-30
Add support for `bytes` parameter type

simonferquel:
Some CNABs might have a need for parameters taking binary data as values. An exemple of such a CNAB would be a generic Helm Chart deployer declaring a HelmChart parameter that should be an actual Tarball with the whole chart in it.

This PR adds support for this scenario and specify what defaultValue / allowedValues should look like for parameters of type `bytes`.

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -349,16 +349,16 @@ Parameter specifications are flat (not tree-like), consisting of name/value pair
 - `parameters`: name/value pairs describing a user-overridable parameter
   - `<name>`: The name of the parameter. This is REQUIRED. In the example above, this is `backend_port`. This
     is mapped to a value definition, which contains the following fields:
-    - `type`: "null", "boolean", "string", "number", or "integer". These types correspond to the main primitive types of JSON (with _integer_ as a special case). Objects and arrays are not supported, as the serialization of these is underdetermined. (REQUIRED)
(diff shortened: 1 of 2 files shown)


technosophos:
How do we represent bytes? Is this something we could do with `string` plus Base 64 encoding? There's also a relevant conversation happening on #77 that might help here (if someone adds support for an array type).

simonferquel:
The parameters would be encoded as base64 in Jason, but injected as a binary file (after decoding the base64 payload)

simonferquel:
Tools might also provide a way to pass the parameters value from a file (thus avoiding base64 encoding)

technosophos:
I'm not sure that you can inject binary data into an environment variable. I think it may contain characters that are forbidden in the env var values of at least some shells. So they'd have to be re-encoded again. I can't seem to find info on what encodings are used for binary env vars.

simonferquel:
Does it make sense to consume binary data as environment vars? Maybe we could restrict binary parameters so that they can only be injected as files?

simonferquel:
Just rebased, sorry for the delay

technosophos:
Nooo.... when I got back to this, another change had caused a conflict! Sorry, @simonferquel .... can I ask you to rebase once more. You can ping me in Slack and I will get it merged right away.

PR #85, opened 2019-01-30
Add description and stateless fields to action

simonferquel:
Custom actions lacked a description field. This fixed that. Also introduces stateless custom actions to implement things like `dry-run`, `help` etc.

The docker-app tool currently abuse the spec to do this kind of things: `inspect` and `render` actions are executed without passing credentials, and are used to visualize the effect of parameter values on the generated stack. We think it should be part of the standard as we see similar requirements for many other generic CNAB (Helm, Terraform, ...)

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -444,10 +444,16 @@ Implementations MAY support user-defined additional actions as well. Such action
 ```json
 "actions": {
     "status":{
-        "modifies": false
+        "modifies": false,
+        "description": "retrieves the status of an installation"
     },
     "migrate":{
         "modifies": false
+    },
+    "dry-run":{
+        "modifies": false,
+        "stateless": true,
+        "description": "prints what install would do with the given parameters values"
     }
 }
 ```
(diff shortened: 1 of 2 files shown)


101-bundle-json.md
```diff
@@ -457,9 +463,14 @@ The above declares to actions: `status` and `migrate`. This means that the assoc
 Each action is accompanied by a description, which contains the following fields:
 
 - `modifies`: Indicates whether the given action will _modify resources_ in any way.
+- `description`: A human readable description of the action (OPTIONAL)
+- `stateless`: The action does not act on a claim, and does not require credentials. This is usefull for exposing dry-run actions, printing documentation, etc. (OPTIONAL)
 
 The `modifies` field MUST be set to `true` if any resource that is managed by the bundle is changed in any way. The `modifies` field assists CNAB implementations in tracking history of changes over time. An implementation of CNAB MAY use this information when describing history or managing releases.
 
+The `stateless` field indicates that the runtime bypass credentials validation (the user MAY or MAY NOT pass any credentials), and will not keep track of this action. Primary intent is to provide Invocation Images to provide dry-run and detailed help functionalities.
```
radu-matei: This is a bit unclear: _intent is to provide Invocation Images to provide_

simonferquel:
Thankis @radu-matei, fixed your comments, but mot sure yet if it is proper english :)

PR #86, opened 2019-01-30
Clarify use of semver

carolynvs:
* Link to the definition of WD from the spec
* Remove text saying that we use semver
* Explain that the version is a _target_ that is applied before we hit it

### DIFF
--- a/901-process.md
+++ b/901-process.md
@@ -4,7 +4,7 @@ While CNAB has not officially been assigned to a standards body, it is good prac
 
 ## Versioning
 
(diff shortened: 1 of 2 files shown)


901-process.md
```diff
@@ -4,7 +4,7 @@ While CNAB has not officially been assigned to a standards body, it is good prac
 
 ## Versioning
 
-Specifications shall each track their own versions, where a version is [semantically versioned](https://semver.org).
+Specifications shall each track their own versions. A specifiction will call out a target version (such as 1.0), and then also include its phase (such as Working Draft, see below) to indicate progress towards the target version.
```
radu-matei: typo: _specifiction _

carolynvs:
Fixed the typo 😊

PR #87, opened 2019-01-31
Add required / optional status for metadata and destination for parameters

radu-matei:
Required status for metadata and destination for parameters was missing from [the parameter spec for bundles](https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md#parameters).

This PR adds the required status for these two properties.


radu-matei:
Also updated the JSON schema, marking the parameter destination as required.

PR #88, opened 2019-02-01
Introduce go types

simonferquel:
This introduce a specs-go golang package, in a similar way as the OCI image-spec repo.
The basic idea is:
- Have a common type system for every golang-written CNAB tool
- Have this common type system kept in sync with the evolution of the specification

### DIFF
--- a/specs-go/v1/bundle.go
+++ b/specs-go/v1/bundle.go
@@ -1,0 +1,327 @@
+package v1
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+// Bundle represent a CNAB descriptor
+type Bundle struct {
+	// The version of the CNAB specification. This should always be the integer 1 for this schema version.
+	SchemaVersion string `json:"schemaVersion"`
+	// The name of this bundle
+	Name string `json:"name"`
+	// A SemVer2 version for this bundle
+	Version string `json:"version"`
+	// A description of this bundle, intended for users
+	Description string `json:"description,omitempty"`
+	// A list of keywords describing the bundle, intended for users
+	Keywords []string `json:"keywords,omitempty"`
+	// A list of parties responsible for this bundle, with contact info
+	Maintainers []Maintainer `json:"maintainers,omitempty"`
+	// The SPDX license code or proprietary license name for this bundle
+	License string `json:"license,omitempty"`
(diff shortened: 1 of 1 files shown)


bacongobbler:
Isn't this the goal of the JSON schema? In theory we should be able to run a code generator to generate this code rather than having to maintain it in two places.

For reference: https://github.com/deislabs/cnab-spec/blob/master/schema/bundle.schema.json

simonferquel:
I don't know any generator capable of dealing correctly with vartyped values (like parameters default value and allowed values).

Also having the strucs actually coming from this repo, would allow common vendoring from one tool to the other and makes it easier to aggregate cnab tools

radu-matei:
However, a related question here is whether to split the relevat  `pkg` part of Duffle into its own Go SDK - and if it does make sense, when to do it in order to keep the friction as low as possible for Duffle?

mattfarina:
@radu-matei go modules does pruning of non-included packages. dep has something similar. I would suggest keeping the `pkg` in duffle for now and let the dependency managers prune the un-needed code.

@simonferquel The neat thing about a spec is all the things that can implement it. Duffle, Docker Apps, and so many other things. I agree with @garethr that this should be in it's own repo. Code and the spec the code implements should live separately.

silvin-lubecki:
What about defining these structs using an more neutral IDL, like [Protocol Buffers](https://developers.google.com/protocol-buffers/) or [Apache Thrift](https://thrift.apache.org/), which are well known by developers and can generate code to almost all the popular languages ?

garethr:
Maintaining multiple computer readable descriptions of the spec manually comes with a cost, and a barrier to entry to spec maintainers. If we want those I'd be tempted to generate them from the JSON Schema rather than see separate bugs in multiple implementations, or wait until the spec is more stable. 

Note that I generated the start of the Python client from the JSON Schema. I'd argue, for it's flaws, JSON Schema is more general than protobufs or Thrift.

technosophos:
Here are some examples of generating Go structs from JSON Schema: https://adrianhesketh.com/2016/07/19/json-schema-to-go-struct-generator-roundup/

I gave this one a try last year, and it seemed to do approximately what I wanted: https://github.com/a-h/generate I think it might work in this case, too.

simonferquel:
No it is not ! Closing :)

PR #91, opened 2019-02-01
rename uri to image

### LINKED ISSUE #89: URI field in image
In some examples there is a `uri` field on `image` that appears (see examples in [101-bundle-json.md](https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md#the-bundlejson-file) and [105-signing.md](https://github.com/deislabs/cnab-spec/blob/master/105-signing.md#digesting-objects-in-the-bundlejson):

```
   "images": [
        {
            "name": "image1",
            "digest": "sha256:aaaa624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b",
            "uri": "urn:image1uri",
            "refs": [
                {
                    "path": "image1path",
                    "field": "image.1.field"
                }
            ],
        }
    ]
```

However, this field appears neither in the spec, nor in the JSON schema.
Should it be removed from the examples, or added in the spec and schema?

bacongobbler:
closes #89

Signed-off-by: Matthew Fisher <matt.fisher@microsoft.com>

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -46,7 +46,6 @@ The following is an example of a `bundle.json` for a bundled distributed as a _t
             "image": "technosophos/microservice:1.2.3",
             "description": "my microservice",
             "digest": "sha256:aaaaaaaaaaaa...",
-            "uri": "urn:image1uri",
             "refs": [
                 {
                     "path": "image1path",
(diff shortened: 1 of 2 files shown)


101-bundle-json.md
```diff
@@ -46,7 +46,7 @@ The following is an example of a `bundle.json` for a bundled distributed as a _t
             "image": "technosophos/microservice:1.2.3",
             "description": "my microservice",
             "digest": "sha256:aaaaaaaaaaaa...",
-            "uri": "urn:image1uri",
+            "image": "image1name:image1tag",
```
radu-matei: This would result in duplicate keys for `image`.

bacongobbler: oh! I missed that. Good catch.

bacongobbler:
It also looks like the spec differs between doc 101 and 105. I'll update 105 to be more in line with the spec.

PR #101, opened 2019-02-11
Introduce namespaced custom actions

silvin-lubecki:
With the multiplication of CNAB runtime implementations, it may occur some custom action clash as multiple runtimes will define different actions using the same name (eg `status`).

I guess it is not be a problem for read-only (`"modifies": false`) actions, but for the others the semantic may change dramatically depending the runtime.

To tackle this problem I suggest we add the following line to the spec:
```
The action name SHOULD be namespaced by the CNAB implementation which generated the bundle.
```
so runtimes will be able to resolve correctly their own custom actions (or if they decide to support custom actions from other runtimes).

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -466,14 +466,14 @@ Implementations MAY support user-defined additional actions as well. Such action
 
 ```json
 "actions": {
-    "status":{
+    "io.cnab.status":{
         "modifies": false,
         "description": "retrieves the status of an installation"
     },
-    "migrate":{
+    "io.cnab.migrate":{
         "modifies": false
     },
-    "dry-run":{
+    "io.cnab.dry-run":{
         "modifies": false,
         "stateless": true,
         "description": "prints what install would do with the given parameters values"
@@ -481,7 +481,9 @@ Implementations MAY support user-defined additional actions as well. Such action
 }
 ```
 
-The above declares to actions: `status` and `migrate`. This means that the associated invocation images can handle requests for `status` and `migrate` in addition to `install`, `upgrade`, and `uninstall`.
+The action _name_ SHOULD be namespaced and SHOULD use reverse DNS notation - e.g. `com.example.action`. 
+
+The above declares three actions: `io.cnab.status`, `io.cnab.migrate` and `io.cnab.dry-run`. This means that the associated invocation images can handle requests for `io.cnab.status`, `io.cnab.migrate` and `io.cnab.dry-run` in addition to `install`, `upgrade`, and `uninstall`.
 
 Each action is accompanied by a description, which contains the following fields:
 
--- a/schema/bundle.schema.json
+++ b/schema/bundle.schema.json
@@ -78,7 +78,7 @@
             }
         },
         "actions":{
-            "description": "Custom actions that can be triggered on this bundle",
+            "description": "Custom actions that can be triggered on this bundle, action name should be namespaced and use reverse DNS notation",
             "type": "object",
             "additionalProperties": {
                 "type": "object",


101-bundle-json.md
```diff
@@ -442,16 +442,18 @@ Implementations MAY support user-defined additional actions as well. Such action
...
     },
-    "migrate":{
+    "io.cnab.migrate":{
         "modifies": false
     }
 }
 ```
 
-The above declares to actions: `status` and `migrate`. This means that the associated invocation images can handle requests for `status` and `migrate` in addition to `install`, `upgrade`, and `uninstall`.
+The action _name_ SHOULD be namespaced by the CNAB implementation which generated the bundle.
```
glyn: This presumes a close relationship between the code which generates the bundle and the runtime which consumes it. I think the namespace should actually be that of the runtime which will recognise the action rather than the code which generates the bundle.

technosophos: As I read it, the suggestion is that the "bundle generator" is the thing that is responsible for adding the namespaced ID, and does not say anything about how that operates. I think Silvin is merely expressing what agent is responsible for _writing the namespaced data_.

101-bundle-json.md
```diff
@@ -442,16 +442,18 @@ Implementations MAY support user-defined additional actions as well. Such action
...
+    "io.cnab.migrate":{
         "modifies": false
     }
 }
 ```
 
-The above declares to actions: `status` and `migrate`. This means that the associated invocation images can handle requests for `status` and `migrate` in addition to `install`, `upgrade`, and `uninstall`.
+The action _name_ SHOULD be namespaced by the CNAB implementation which generated the bundle. 
+
+The above declares two actions: `io.cnab.status` and `io.cnab.migrate`. This means that the associated invocation images can handle requests for `status` and `migrate` in addition to `install`, `upgrade`, and `uninstall`.
```
glyn: There needs to be a normative definition of namespacing. [Reverse domain names](https://en.wikipedia.org/wiki/Reverse_domain_name_notation) are reasonable, but the spec needs to be clear so that users don't choose varying namespacing schemes.

technosophos: I believe OCI specifies these SHOULD use reverse domains (https://github.com/opencontainers/image-spec/blob/master/annotations.md). But I think @mattfarina was in favor of staying less stringent and just suggesting "unambiguous dotted namespaces".

silvin-lubecki: That's true, but as namespacing is non mandatory (`SHOULD`), I don't if we can enforce a particular schema, other than just saying "Reverse domain names is preferred".  WDYT?

101-bundle-json.md
silvin-lubecki: I think the "bundle generator" is more aware of what actions it put into the invocation image and how it should be run, more than the "bundle runtime".

garethr: I think if you remove "by the CNAB implementation which generated the bundle." the meaning of this doesn't change and the spec has less words in it. I think the meaning of "CNAB implementation" is up for debate and removing it removes the potential for ambiguity.

silvin-lubecki: Ok, let me remove this part then 👍

101-bundle-json.md
technosophos: I agree.

silvin-lubecki:
Just rebased on top of master, integrating new `dry-run` custom action example introduced by @simonferquel, added SHOULD `use reverse domains notation` and updated the `bundle.schema.json` file with it.
PTAL @technosophos @glyn @garethr

101-bundle-json.md
```diff
@@ -442,22 +442,24 @@ Implementations MAY support user-defined additional actions as well. Such action
...
         "stateless": true,
         "description": "prints what install would do with the given parameters values"
     }
 }
 ```
 
-The above declares to actions: `status` and `migrate`. This means that the associated invocation images can handle requests for `status` and `migrate` in addition to `install`, `upgrade`, and `uninstall`.
+The action _name_ SHOULD be namespaced and SHOULD use reverse domains notation - e.g. `com.example.myKey`. 
+
+The above declares three actions: `io.cnab.status`, `io.cnab.migrate` and `io.cnab.dry-run`. This means that the associated invocation images can handle requests for `status`, `migrate` and `dry-run` in addition to `install`, `upgrade`, and `uninstall`.
```
glyn: If invocation images are passed truncated action names (e.g. `status`) rather than fully-qualified action names (e.g. `io.cnab.status`), there could be collisions. Why not pass the fully-qualified action names to invocation images instead?

101-bundle-json.md
mattfarina: I like the idea of reverse domain names being preferred. I can update the namespacing PR to reflect that.

mattfarina:
I'm a little confused about the purpose of namespacing the actions. What problem does it solve?

For example, if I have one tool create the bundle shouldn't other tools that try to execute it be able to perform/call the actions?

At a high level I'm wondering how to handle multi-tool compatibility. What is the idea around that? This is a little different from namespacing end user naming of things in the bundle json (which might be a bit overcome by events with the OCI repos, right?).

101-bundle-json.md
technosophos: And we should at least be specific about what would happen if the action was named `io.cnab.this.is.a.long.name`: is it just the last string (`name`) or `this.is.a.long.name`?

technosophos:
@mattfarina I think the general idea is that a single verb (`status`) is too common, and may lead to confusion when two different packages accidentally use the same verb, whereas `io.cnab.status` is more specific.

I think that as with OCI annotations and Kubernetes labels, we could definitely write little microformats for particular ones (like `io.cnab.status`).

101-bundle-json.md
technosophos: @mattfarina Okay. That would be good... or did I merge that one already?

101-bundle-json.md
glyn: Why do you want to treat the prefix `io.cnab` specially? Is it because you'd like standard action names to have fully-qualified synonyms? For example, would you like `io.cnab.install` to be a synonym for `install`?

I don't really think this is necessary because actions with a single component, e.g. `install`, are clearly distinct from reverse domain names. (However, it would be a bad idea to name a standard action after a top-level domain name such as `org` as that would then also be a reverse domain name!)

So, if you agree that we can use single component names for standard actions, then we can treat the above example as a regular custom name: `io.cnab.this.is.a.long.name` would not have a short form. In other words, we wouldn't need to treat the prefix `io.cnab` specially.

garethr: +1 on @glyn point. I don't think the special casing is worth the extra effort in either the spec to describe the edge cases or in making all clients more complicated to deal with that logic.

I would remove this paragraph.

I think the rationale for doing so would be user interface, but that's a client concern, not for the spec.

101-bundle-json.md
```diff
@@ -442,22 +442,24 @@ Implementations MAY support user-defined additional actions as well. Such action
...
+    "io.cnab.dry-run":{
         "modifies": false,
         "stateless": true,
         "description": "prints what install would do with the given parameters values"
     }
 }
 ```
 
-The above declares to actions: `status` and `migrate`. This means that the associated invocation images can handle requests for `status` and `migrate` in addition to `install`, `upgrade`, and `uninstall`.
+The action _name_ SHOULD be namespaced and SHOULD use reverse domains notation - e.g. `com.example.myKey`.
```
garethr: Slight incongruity with the examples using hyphen-separators and the e.g. here using compoundWords. The spec is neutral about which to use, but [RFC4343 ](https://tools.ietf.org/html/rfc4343 ) states _Domain Name System (DNS) names are "case insensitive"_.

schema/bundle.schema.json
```diff
@@ -78,7 +78,7 @@
             }
         },
         "actions":{
-            "description": "Custom actions that can be triggered on this bundle",
+            "description": "Custom actions that can be triggered on this bundle, action name should be namespaced and use reverse domains notation",
```
garethr: `domains` should be DNS.

101-bundle-json.md
garethr: `domains` should be DNS

simonferquel:
I agree that passing full action names should be better. Especially when a single cnab wants to play nice with different tools that relies on namespaced actions.
For example both tool-a and tool-b could define their own well-known status action as tool-a.status and tool-b.status, expecting different output. A same cnab should be able to play well with both.

101-bundle-json.md
silvin-lubecki: I'm fine with passing fully-qualified action names, any objection @technosophos @garethr ?

101-bundle-json.md
silvin-lubecki: Fixed

schema/bundle.schema.json
silvin-lubecki: Fixed

101-bundle-json.md
silvin-lubecki: Renamed to `com.example.action`

silvin-lubecki:
Replaced `domains` with `DNS`, changed the reverse DNS example.

>The above declares three actions: `io.cnab.status`, `io.cnab.migrate` and `io.cnab.dry-run`. This means that the associated invocation images can handle requests for `status`, `migrate` and `dry-run` in addition to `install`, `upgrade`, and `uninstall`.

@garethr you proposed to remove this paragraph, but I think it clarifies the fact the invocation image still needs to handle mandatory actions `install`, `upgrade` and `uninstall` and the custom actions don't supersede them.

glyn:
> Replaced `domains` with `DNS`, changed the reverse DNS example.
> 
> > The above declares three actions: `io.cnab.status`, `io.cnab.migrate` and `io.cnab.dry-run`. This means that the associated invocation images can handle requests for `status`, `migrate` and `dry-run` in addition to `install`, `upgrade`, and `uninstall`.
> 
> @garethr you proposed to remove this paragraph, but I think it clarifies the fact the invocation image still needs to handle mandatory actions `install`, `upgrade` and `uninstall` and the custom actions don't supersede them.

If you keep the paragraph, please change it to use the fully-qualified action names throughout.

silvin-lubecki:
> The above declares three actions: `io.cnab.status`, `io.cnab.migrate` and `io.cnab.dry-run`.

I already did 👍

garethr:
+1 from me. I think my comment referred to a previous version which has been markedly improved.

PR #102, opened 2019-02-12
Introduce well known custom actions

simonferquel:
This list a set of well known actions that a bundle MAY implement, and that tools can understand and interact with, possibly with a premium UI.

Current list is 
- `io.cnab.help` (with `stateless`: true and `modifies`: false): print an help message to the standard output. Implementations MAY print different messages depending on the parameters values passed to the invocation image.
- `io.cnab.log` (with `stateless`: false and `modifies`: false): print logs of the installed system to the standard output.
- `io.cnab.status` (with `stateless`: false and `modifies`: false): print a human readable status message to the standard output.
- `io.cnab.status+json` (with `stateless`: false and `modifies`: false): print a json payload describing the detailed status using a well known schema.


103-bundle-runtime.md
```diff
@@ -84,6 +84,59 @@ In addition to the default actions, CNAB runtimes MAY support custom actions (as
 
 A bundle MUST exit with an error if the action is executed, but fails to run to completion. A CNAB runtime MUST issue an error if a bundle issues an error. And an error MUST NOT be issued if one of the three built-in actions is requested, but not present in the bundle. Errors are reserved for cases where something has gone wrong.
 
+#### Well known custom actions
+
+An invocation image MAY implement any of the following well known custom actions. A CNAB indicates that it supports those actions by including them in its custom action list (as defined in [the bundle definition](101-bundle-json.md)).
+- `io.cnab.help` (with `stateless`: true and `modifies`: false): print an help message to the standard output. Implementations MAY print different messages depending on the parameters values passed to the invocation image.
+- `io.cnab.log` (with `stateless`: false and `modifies`: false): print logs of the installed system to the standard output.
```
ijc: Is this is a one shot dump of whatever the current set of logs, potentially missing the start due to rotation etc? Or is this expected to have `tail -f` type behaviour?

103-bundle-runtime.md
```diff
@@ -84,6 +84,59 @@ In addition to the default actions, CNAB runtimes MAY support custom actions (as
...
+- `io.cnab.status` (with `stateless`: false and `modifies`: false): print a human readable status message to the standard output.
+- `io.cnab.status+json` (with `stateless`: false and `modifies`: false): print a json payload describing the detailed status with the following schema:
+
+```json
+{
+    "status": "Failed",
+    "message": "Component frontend failed to deploy properly",
```
ijc: Nit, the component is called `front-end` below. I suppose an implementation is at liberty to be inconsistent but maybe not to be encouraged.

103-bundle-runtime.md
```diff
@@ -84,6 +84,59 @@ In addition to the default actions, CNAB runtimes MAY support custom actions (as
...
+        "front-end": {
+            "status": "Failed",
+            "message": "Failed to pull Docker image mginx:latest",
+        },
+    }
+}
+```
+  - `status` is a value indicating the deployment status of a component or of the whole bundle. Well known values are `Failed`, `Ready`, `Pending`. An invocation image MAY use other custom-values, but tools must at least understand those values.
+  - `message` optional field giving details about the current bundle or component status.
+  - `components`: map of components/statuses allowing a more detailed status. A component can itself have subcomponent, to enable scenarios like aggregating statuses in CNABs composed of other CNABs.
```
ijc: Perhaps an example of the "subcomponent" case would be good, in particular to confirm whether the expected field name is `"components"` vs `"subcomponents"`.

nit: " can itself have subcomponent" should end "subcomponents".

103-bundle-runtime.md
```diff
@@ -84,6 +84,59 @@ In addition to the default actions, CNAB runtimes MAY support custom actions (as
...
+  - Additionaly to the fields defined above, an invocation image MAY add custom fields as they want. An example of that could be a CNAB bundle deploying Kubernetes workloads and exposing scale details:
```
ijc: At the risk of recursing: should these custom fields be namespaced ala `com.foo.bar`?

103-bundle-runtime.md
simonferquel: Initially, it will be one shot of the current set logs. We are preparing a proposal to scope parameters to a given set of actions. If it passes, I'll update this one with a `io.cnab.log.follow` parameter scoped to the `io.cnab.log` action for allowing `-f`-like behavior.

103-bundle-runtime.md
simonferquel: We can suggest that in the spec without enforcing this, yes.

103-bundle-runtime.md
```diff
@@ -84,6 +84,83 @@ In addition to the default actions, CNAB runtimes MAY support custom actions (as
...
+#### Well known custom actions
+
+An invocation image MAY implement any of the following well known custom actions. A CNAB indicates that it supports those actions by including them in its custom action list (as defined in [the bundle definition](101-bundle-json.md)).
+- `io.cnab.help` (with `stateless`: true and `modifies`: false): print an help message to the standard output. Implementations MAY print different messages depending on the parameters values passed to the invocation image.
+- `io.cnab.log` (with `stateless`: false and `modifies`: false): print logs of the installed system to the standard output.
+- `io.cnab.status` (with `stateless`: false and `modifies`: false): print a human readable status message to the standard output.
+- `io.cnab.status+json` (with `stateless`: false and `modifies`: false): print a json payload describing the detailed status with the following schema:
+
+```json
+{
```
silvin-lubecki: I think this should also be backed by a `json schema`, which is less subject to interpretation and which could be used by runtimes to validate the output before rendering it (UI, standard output,...).

103-bundle-runtime.md
```diff
@@ -84,6 +84,83 @@ In addition to the default actions, CNAB runtimes MAY support custom actions (as
...
+                }
+            }
+        },
+        "wordpress": {
+            "status": "Ready"
+        }
+    }
+}
+```
+  - Additionaly to the fields defined above, an invocation image MAY add custom fields as they want. To avoid amiguous naming, those fields names MAY be namespaced. An example of that could be a CNAB bundle deploying Kubernetes workloads and exposing scale details:
```
silvin-lubecki: typo: ambiguous

garethr:
I think I'd prefer to see things like this arise outside the spec itself in the user community. Specifying things before any implementation is high risk. Better to see patterns emerge from successful usage and to to see experiments in tools first, and specification later. The spec should be as minimal as possible, I feel like SHOULD items without implementation should be avoided..

technosophos:
Should we just move it to an 800-level (non-normative) part of the spec, @garethr ?

garethr:
@technosophos that works for me. Keeping things in one place is good to spur conversations, but easier to evolve. If things are widely adopted there by clients in particular it's easy to move them into the spec proper.

simonferquel:
I will move that to a separate 800-level md file then

PR #103, opened 2019-02-12
Intoduce action-specific parameters

simonferquel:
This adds an optional "apply-to" field to parameters definition to make
a parameter usable only for a specific set of actions.

eg.: we could have a `follow` parameter that could only apply to the
`logs` custom action of a CNAB (as it only makes sense in the context of
this action)

### DIFF
--- a/schema/bundle.schema.json
+++ b/schema/bundle.schema.json
@@ -308,6 +308,10 @@
                             "type": "string"
                         }
                     }
+                },
+                "apply-to":{                    
+                    "description": "An optional exhaustive list of actions handling this parameter",
+                    "type":"array"
                 }
             },
             "required": ["type", "destination"]
--- a/schema/bundle.schema.json
+++ b/schema/bundle.schema.json
@@ -311,7 +311,10 @@
                 },
                 "apply-to":{                    
                     "description": "An optional exhaustive list of actions handling this parameter",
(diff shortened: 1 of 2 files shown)


schema/bundle.schema.json
```diff
@@ -308,6 +308,10 @@
                             "type": "string"
                         }
                     }
+                },
+                "apply-to":{                    
+                    "description": "An optional exhaustive list of actions handling this parameter",
+                    "type":"array"
```
garethr: This should probably also specify that the items can only be strings, given the purpose is a list of actions. Anything else should be invalid I think.

```json
"items": {
  "type": "string"
}
```

technosophos:
I have to think through the use cases a bit, but at first blush this makes sense.

carolynvs:
I like this idea from the "don't prompt me for silly parameters that aren't relevant to this action" standpoint. If only install needs X, then if I'm doing a status, don't force me to specify X. 👍

simonferquel:
I thought more in the other way: I have a log acton that has a "follow" parameter. Don't make me fill it on install.
But is suppose both apply.

technosophos:
Is this one ready to merge? It looks like it.

simonferquel:
Feels like it (very small change IMO, with great UX improvement potential)

PR #104, opened 2019-02-12
Bundle extension fields through `custom`

technosophos:
This PR supersedes #28 

This adds a `custom` key as a designated place where extensions may be inserted into the `bundle.json` file.

See #28 for relevant discussion.

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -502,22 +502,22 @@ The built-in actions (`install`, `upgrade`, `uninstall`) MUST NOT appear in the 
 
 Implementations that do not support custom actions MUST NOT emit errors (either runtime or validation) if a bundle defines custom actions. That is, even if an implementation cannot execute custom actions, it MUST NOT fail to operate on bundles that declare custom actions.
 
-## Extensions
+## Custom Extensions
 
-In many cases, the bundle descriptor is a sufficient artifact for delivering a CNAB bundle, since invocation images and other images may be retrieved from registries and repositories. Consequently, it is important to provide an extension mechanism. An _extension_ is a named collection of auxiliary data whose meaning is defined outside of this specification.
+In many cases, the bundle descriptor is a sufficient artifact for delivering a CNAB bundle, since invocation images and other images may be retrieved from registries and repositories. Consequently, it is important to provide an extension mechanism. A _custom extension_ is a named collection of auxiliary data whose meaning is defined outside of this specification.
 
-Tools MAY define and declare additional fields inside of the `extensions` section. Tools MUST NOT define additional fields anywhere else in the bundle descriptor. Implementations MAY produce an error or MAY ignore additional fields outside of the extension. However, implementations SHOULD preserve the data inside of the `extensions` section even when that information is not understood by the implementation.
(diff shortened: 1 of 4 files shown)


technosophos:
This is mostly the same as #28, except it uses `custom` instead of `extensions`

101-bundle-json.md
```diff
@@ -500,4 +506,45 @@ The built-in actions (`install`, `upgrade`, `uninstall`) MUST NOT appear in the
 
 Implementations that do not support custom actions MUST NOT emit errors (either runtime or validation) if a bundle defines custom actions. That is, even if an implementation cannot execute custom actions, it MUST NOT fail to operate on bundles that declare custom actions.
 
+## Custom Extensions
+
+In many cases, the bundle descriptor is a sufficient artifact for delivering a CNAB bundle, since invocation images and other images may be retrieved from registries and repositories. Consequently, it is important to provide an extension mechanism. A _custom extension_ is a named collection of auxiliary data whose meaning is defined outside of this specification.
```
glyn: Maybe "However" would be more appropriate than "Consequently" unless the previous sentence gave an example which _isn't_ covered by just the standard bundle descriptor data.

101-bundle-json.md
```diff
@@ -500,4 +506,45 @@ The built-in actions (`install`, `upgrade`, `uninstall`) MUST NOT appear in the
...
+
+Tools MAY define and declare additional fields inside of the `custom` section. Tools MUST NOT define additional fields anywhere else in the bundle descriptor. Implementations may produce an error or may ignore additional fields outside of the extension. However, implementations SHOULD preserve the data inside of the `custom` section even when that information is not understood by the implementation.
```
glyn: Reword third sentence to: "Implementations may ignore additional fields outside of the extension or may produce an error when such fields are present."

silvin-lubecki: nit: MAY in upper case

101-bundle-json.md
```diff
@@ -500,4 +506,45 @@ The built-in actions (`install`, `upgrade`, `uninstall`) MUST NOT appear in the
...
+    "extensions": {
+        "EXTENSION NAME": "ARBITRARY JSON DATA"
+    }
+}
+```
+
+The fields are defined as follows:
+
+- `extensions` defines the wrapper object for extensions
+  - EXTENSION NAME: a unique name for an extension. Names SHOULD follow the reverse DNS format `DOMAIN.ADDITIONAL.INFORMATION`, where DOMAIN conforms to the reverse DNS naming convention (`com.example` or `io.cnab`), and additional elements are separated by dots (`.`).
```
silvin-lubecki: Should we factorize this part so we can reference it wherever we need namespaced names? (custom actions, labels, extensions...)

101-bundle-json.md
technosophos: Fixed

101-bundle-json.md
technosophos: Reworded to explain that this is an exclusive disjunction.

technosophos: Fixed

101-bundle-json.md
technosophos: Done.

PR #110, opened 2019-02-15
fix: remove CNAB_P_* support

### LINKED ISSUE #105: Removing support for CNAB_P_<param>
Should we remove support for `CNAB_P_<param name>`?

Early CNAB did not allow custom parameter paths or env vars, and instead automatically injected parameters according the the formula `CNAB_P_<PARAM NAMe>` (e.g. `CNAB_P_PORT`)

Later, we added support for `destination`, which allows setting the desired destination.

Is it time to remove the `CNAB_P_` style parameters, or should we leave them?

technosophos:
This removes the autogeneration of CNAB_P_* environment variables, and requires the `destination` field on parameters.

Additionally, this change provides a little more clarification on how `path` is to be handled in `destination`.

Closes #105

### DIFF
--- a/103-bundle-runtime.md
+++ b/103-bundle-runtime.md
@@ -84,11 +84,11 @@ In addition to the default actions, CNAB runtimes MAY support custom actions (as
 
 A bundle MUST exit with an error if the action is executed, but fails to run to completion. A CNAB runtime MUST issue an error if a bundle issues an error. And an error MUST NOT be issued if one of the three built-in actions is requested, but not present in the bundle. Errors are reserved for cases where something has gone wrong.
 
-## Overriding Parameter Values
+## Setting Parameter Values
 
 A CNAB `bundle.json` file MAY specify zero or more parameters whose values MAY be specified by a user.
 
-As specified, values MAY be passed into the container as environment variables. If the environment variable name is specified in the `destination`, that name will be used:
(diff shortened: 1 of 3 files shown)


103-bundle-runtime.md
```diff
@@ -105,29 +105,40 @@ As specified, values MAY be passed into the container as environment variables.
...
 
-The above will set `GREETING=hello`.
+By default (if no override value is provided by the CNAB runtime), the above will set `GREETING=hello`. If the runtime specifies a value `salutations`, then the environment variable would be set to `GREETING=salutations`.
 
-In the case where no `destination` is set, a parameter is written as an environment variable with an automatically generated name.
+The parameter value is evaluated thus:
+
+- If the CNAB runtime provides a value, that value MAY be sanitized, then validated (as described below), then injected as the parameter value. In the event that sanitization or validation fail, the runtime SHOULD return an error and discontinue the action.
+- If the CNAB runtime does not provide a value, but `defaultValue` is set, then the default value MUST be used.
+- If no value is provided and `defaultValue` is unset, the runtime MUST set the value to an empty string (""), regardless of type.
```
silvin-lubecki: What about adding a paragraph on `required` parameter?
```
If the parameter has a field `required` set to `true`, then the CNAB runtime MUST return an error and discontinue the action.
```

PR #111, opened 2019-02-15
feat(ci): add validation automation

vdice:
Implements https://github.com/deislabs/cnab-spec/issues/109

Adds:

- `validate.sh`, a schema validation script for vetting all bundle.json examples in `examples` against the schema files in `schema`

- `Makefile` with validation targets invoking said script, where 
  - `make validate` defaults to a Docker-based validation approach
  - `make validate-local` runs locally
  - (open to switching these such that `validate` runs locally and `validate-docker` runs in the Docker image, if desired)

- `brigade.js` for running this validation on pull request and `master` branch push events

### DIFF
--- a/Dockerfile.ajv
+++ b/Dockerfile.ajv
@@ -1,0 +1,3 @@
+FROM node:8-alpine
+
+RUN npm install -g ajv-cli
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,23 @@
+BASE_DIR         := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+VALIDATOR_IMG    := deislabs/cnab-spec.ajv
+VALIDATOR_SCRIPT := ./validate.sh
+
+.PHONY: build-validator
+build-validator:
+	@docker build -f Dockerfile.ajv -t $(VALIDATOR_IMG) .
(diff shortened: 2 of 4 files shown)


garethr:
I think this should be part of the examples repository, rather than part of the spec. I think mixing tools and the spec is a slippery slope, given everything from different change control processes to different versioning, etc.

I'd caution as well that the schema can't describe everything in the spec, JSON Schema isn't quite powerful enough and focuses just on the bundle.json.

As mentioned in Slack I'm not advocating for _my_ approach, rather any one that covers all bases. I demonstrated validating the current examples using the Python client here: https://gist.github.com/garethr/0b30883e3da92ad19d89a5874e8b295a

vdice:
Thanks @garethr.

In fact, the validation added here originates in the validation in example-bundles: https://github.com/deislabs/example-bundles/runs/62993823 (😬 yup, these bundles still need some TLC). 

I believe we may still want this added validation in this repo to ensure the examples in the `examples` directory are aligned with the schemas present.  For instance, the validation uncovered the following update PR to said examples: https://github.com/deislabs/cnab-spec/pull/112

garethr:
Yup, works for me.

PR #112, opened 2019-02-15
fix(examples): update examples with required parameter destination

vdice:
Updates example bundle.json files in `examples` to specify the required parameter destination where applicable.

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -71,6 +71,9 @@ The following is an example of a `bundle.json` for a bundled distributed as a _t
   "parameters": {
     "backend_port": {
       "defaultValue": 80,
+      "destination": {
+        "env": "BACKEND_PORT"
+      },
       "maxValue": 10240,
       "metadata": {
         "description": "The port that the back-end will listen on"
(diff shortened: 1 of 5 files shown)


vdice:
@technosophos I just realized that there are some doc files (`101-bundle-json.md`, etc.) that reference the example bundle.json contents.  A few questions:

1. It appears as of today it is a manual effort to re-copy updated examples into the appropriate locations of the docs?
2. I see some instances of corresponding canonical json representations, which I'm assuming should also be updated -- what tool was used to generate these representations?
3. Thoughts on how we might avoid the manual effort of 1 & 2 (if I have those right)?

technosophos:
1. It is currently a manual effort. 🐼 
2. I left a note somewhere in 101 saying `Small snippets of JSON may be shown in the order in which the fields are described (for clarity) rather than in Canonical JSON order.` -- mainly because I didn't think I could catch and correct all of the fields. Ideally, though, they should be in canonical JSON
3. I guess we could use a preprocessor to insert them. It would probably be good to see how other places do this.

vdice:
Last commit updated the json snippets in corresponding docs; PTAL @technosophos

101-bundle-json.md
```diff
@@ -88,67 +91,70 @@ Source: [101.01-bundle.json](examples/101.01-bundle.json)
...
-        "digest": "sha256:bbbbbbbbbbbb...",
-        "image": "technosophos/helloworld:0.1.2",
-        "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
-        "platform": {
-          "architecture": "amd64",
-          "os": "linux"
-        },
-        "size": 1337
-      }
+  "credentials": {
```
vdice: indentation fix here... looked like original had bulk of json body outdented by 2 spaces.

PR #117, opened 2019-02-26
fix: render the image map stuff consistent

### LINKED ISSUE #113: Address redundancy in spec regarding declarative image references and runtime references
See https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md#the-image-map. Refs and image map injection at runtime are redundant  and we should clarify the specification to remove the redundancy. 

@simonferquel @lachie83 @technosophos

technosophos:
This fixes some conflicts between 101 and 103 regarding how image mapping is to be done.

Closes #113

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -268,37 +268,47 @@ Fields:
   - `imageType`: The `imageType` field MUST describe the format of the image. The list of formats is open-ended, but any CNAB-compliant system MUST implement `docker` and `oci`. The default is `oci`.
   - `image`: The REQUIRED `image` field provides a valid reference (REGISTRY/NAME:TAG) for the image. Note that SHOULD be a CAS SHA, not a version tag as in the example above.
   - `digest`: MUST contain a digest, in [OCI format](https://github.com/opencontainers/image-spec/blob/master/descriptor.md#digests), to be used to compute the integrity of the image. The calculation of how the image matches the digest is dependent upon image type. (OCI, for example, uses a Merkle tree while VM images are checksums.)
-  - `refs`: An array listing the locations which refer to this image, and whose values should be replaced by the value specified in URI. Each entry contains the following properties:
+  - `refs`: An array listing the locations which refer to this image, and whose values should be replaced by the value specified in URI. The behavior associated with this section is defined below. Each entry contains the following properties:
     - `path`: The path of the file where the value should be replaced
     - `field`: A selector specifying a location (or locations) within that file where the value should be replaced
     - `mediaType`: The media type of the file, which can be used to determine the file type. If unset, tooling may choose any strategy for detecting format
+    - `expressionType`: The expression language that should be used to evaluate `field`.
   - `size`: The image size in bytes
   - `platform`: The target platform, as an object with two fields:
     - `architecture`: The architecture of the image (`i386`, `amd64`, `arm32`...)
     - `os`: The operating system of the image
   - `mediaType`: The media type of the image
 
-Substitutions MUST be supported for the following formats:
+### Reference Substitution (`refs`)
 
-- JSON
-- YAML
-- XML
+The `refs` section provides tools for mapping fully qualified image names into paths present within the invocation image.
 
-In addition to these substitutions, the image map data is also made available to the invocation image at runtime. See [Image map](103-bundle-runtime.md#image-map) for more details.
+A CNAB Runtime MUST pass the images map into the invocation image, as described in [the Bundle Runtime definition](103-bundle-runtime.md#image-map). Invocation images MAY implement support for this feature.
 
-### Field Selectors
+For example, consider the following `refs` declaration:
 
-*TODO:* We have multiple competing standards in this space, and those that are popular for JSON are not the same as those popular for XML. This portion is thus not complete.
+```json
+"images": {
+    "my-microservice": {
+      "description": "my microservice",
+      "digest": "sha256:aaaaaaaaaaaa...",
+      "image": "technosophos/microservice:1.2.3",
+      "refs": [
+        {
+          "field": "service.image",
+          "path": "/cnab/charts/values.yaml",
+          "expressionType": `css-selector`
+        }
+      ]
+    }
+  },
+```
 
-For fields, the selectors are based on the _de facto_ format used in tools like `jq`, which is a subset of the [CSS selector](https://www.w3.org/TR/selectors-3/) path. Examples:
+For the image described as `my-microservice`, the file `/cnab/charts/values.yaml` inside of the invocation image is to be parsed, and inside of that file, the value of `service.image` is to be set to `technosophos/microservice:1.2.3`.
 
-- `foo.bar.baz` is interpreted as "find element baz whose parent is bar and whose grandparent is foo".
-- `#baz` in XML is "the element whose ID attribute is set to "baz"". It is a no-op in YAML and JSON.
-- TODO: Will we need to support attribute selectors?
+The behavior of the invocation image is described in [the Bundle Runtime definition](103-bundle-runtime.md#image-map).
 
-TODO: How do we specify multiple replacements within a single file?
-
-TODO: How do we specify URI is a VM image (or Jar or other) instead of a Docker-style image? Or do we? And if not, why not?
+CNAB Runtimes MAY dynamically create or override `refs` sections when appropriate. For example, if an image is relocated as part of a CNAB installation, the runtime may create a `refs` section to indicate that change.
 
 ## Parameters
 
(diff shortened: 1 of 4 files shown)


technosophos:
/cc @simonferquel @garethr @jeremyrickard @lachie83 

This PR tries to fix the ambiguities mentioned in #113. However, I'm unhappy with my attempt to fix it. The reason is that it pushes the burden on to the bundle developer. This almost seems unavoidable, though, since any other work-around would require the CNAB Runtime to know how to alter and save any image format. That would be an utter nightmare.

Also, there is a statement in @simonferquel 's section that says:

> The run tool MAY use this file to modify its behavior, if declarative substitution is not enough.

I am not sure where `declarative substitution` is defined or what that phrase means.

At the end of the day, I feel like this represents something that _can work_, but which is not ideal. I tried to devise another way of doing this that would be simpler, but I cannot think of anything.

So... this might be incrementally better than what was there before, but it's not a complete solution.

101-bundle-json.md
```diff
@@ -268,37 +268,47 @@ Fields:
...
-- `foo.bar.baz` is interpreted as "find element baz whose parent is bar and whose grandparent is foo".
-- `#baz` in XML is "the element whose ID attribute is set to "baz"". It is a no-op in YAML and JSON.
-- TODO: Will we need to support attribute selectors?
+For the image described as `my-microservice`, the file `/cnab/charts/values.yaml` inside of the invocation image is to be parsed, and inside of that file, the value of `service.image` is to be set to `technosophos/microservice:1.2.3`.
 
-TODO: How do we specify multiple replacements within a single file?
+The behavior of the invocation image is described in [the Bundle Runtime definition](103-bundle-runtime.md#image-map).
 
-TODO: How do we specify URI is a VM image (or Jar or other) instead of a Docker-style image? Or do we? And if not, why not?
+CNAB Runtimes MAY dynamically create or override `refs` sections when appropriate. For example, if an image is relocated as part of a CNAB installation, the runtime may create a `refs` section to indicate that change.
```
glyn: @technosophos Actually, I see image relocation as a responsibility of an invocation image rather than something runtimes need to be concerned about. (From that perspective, the `refs` section isn't needed in the spec.)

101-bundle-json.md
```diff
@@ -268,37 +268,47 @@ Fields:
...
+"images": {
+    "my-microservice": {
+      "description": "my microservice",
+      "digest": "sha256:aaaaaaaaaaaa...",
+      "image": "technosophos/microservice:1.2.3",
+      "refs": [
+        {
+          "field": "service.image",
+          "path": "/cnab/charts/values.yaml",
+          "expressionType": `css-selector`
```
vdice: I was slightly confused to see the expressionType of `css-selector` in this example for setting a chart value.  Was this intentional?  Or should it be `pcre`?

101-bundle-json.md
technosophos: To this point in time, we've assumed the following:

- Invocation images are never responsible for image management (partly because they may not even know the image type (eg. VM, OCI, MSIX, etc)
- Information about an image location must be injected into the bundle from outside

That raised the problem that we're trying to solve here. An example goes like this:

> I have a bundle that contains a Kubernetes manifest. That manifest points to `example.com/foo/bar:v1`, but I just exported the image and imported it into an isolated network. The new image was loaded to `inner-network.local/foo/bar:v1`

The thing that knows about `inner-network.local` is the CNAB Runtime. The invocation image is blissfully unaware of whether it's on the public network or the inner network, and blissfully unaware of its network constraints. It will work just as well in the external public network as in the internal network.

BUT... when we moved the images, we now need to notify the invocation image that it needs to rewrite its Kubernetes manifest to point to the new `inner-network.local/foo/bar:v1` location.

Now, you raise a really good point: If the invocation image knows its own internal file system layout (which it does), then the `refs` field is not particularly useful. The invocation image itself should have all the internal logic it needs. So I suppose we could just ax that part of the spec. What we would seem to need in this case is just a mapping of old image names to their new image names. /cc @simonferquel  <-- is that what you had in mind in your earlier PR?

glyn: > To this point in time, we've assumed the following:
> 
> * Invocation images are never responsible for image management (partly because they may not even know the image type (eg. VM, OCI, MSIX, etc)

I'd like to challenge this assumption.

I wonder if invocation images might actually be _better_ able to support the image types in their bundles than a runtime, which would need a pretty complex extension mechanism to cope with _arbitrary_ image types.

This would avoid the need for explicit support for image management in the spec or runtime.

We could reduce the complexity for bundle authors, by building management features for docker (and other types of) images into base invocation images and/or adding tooling features (such as porter mixins).

> * Information about an image location must be injected into the bundle from outside

I'd also like to challenge this assumption.

In the use case of image relocation, it would be sufficient to pass a parameter describing the target registry. This would avoid the need for explicit support in the spec or runtime.

> 
> That raised the problem that we're trying to solve here. An example goes like this:
> 
> > I have a bundle that contains a Kubernetes manifest. That manifest points to `example.com/foo/bar:v1`, but I just exported the image and imported it into an isolated network. The new image was loaded to `inner-network.local/foo/bar:v1`
> 
> The thing that knows about `inner-network.local` is the CNAB Runtime. The invocation image is blissfully unaware of whether it's on the public network or the inner network, and blissfully unaware of its network constraints. It will work just as well in the external public network as in the internal network.
> 
> BUT... when we moved the images, we now need to notify the invocation image that it needs to rewrite its Kubernetes manifest to point to the new `inner-network.local/foo/bar:v1` location.
> 
> Now, you raise a really good point: If the invocation image knows its own internal file system layout (which it does), then the `refs` field is not particularly useful. The invocation image itself should have all the internal logic it needs. So I suppose we could just ax that part of the spec.

I agree we could ax that part of the spec.

I'd also like us to consider the option of images being provided to a bundle separately to the invocation image: this is the subject of https://github.com/deislabs/cnab-spec/issues/95.

> What we would seem to need in this case is just a mapping of old image names to their new image names. /cc @simonferquel <-- is that what you had in mind in your earlier PR?

Of course, if we made invocation images responsible for image relocation and passed a target registry as a parameter, then the spec and runtime wouldn't need to provide a mapping.

[An aside: the more features we can omit from v1 of the specification, the easier it will be to finalise the specification and implement runtimes.

Additional spec features may turn out to be desirable as the community gains experience in building runtimes for v1, but I think that's a healthy way to discover spec requirements. Trying to nail down all the spec requirements in v1 is a tall order given that only a few runtimes are currently being developed.

In general, it would be much easier to add features to future spec versions than have to cope with features that later turned out to be unnecessary.]

simonferquel: To my understanding, `refs` "declarative" replacements were to be handled by the CNAB runtime, whereas invocation images may support an imperative approach (by consuming the injected image map)

That is, after creation of the container, but before starting it, the driver would apply the `refs` rules without requiring anything within the invocation image.

In my mind, the refs are completely ignored by the invocation image then.

simonferquel:
After a lot of thought, I think the `refs` should actually be dropped: it would make the CNAB runtime drivers really complex if we want to implement declarative replacements handling on the runtime side, and as the invocation image itself has the knowledge of its internal layout, it does not really bring any value if we handle them within the invocation image.

If you look at the bundles created by docker-app, you will see that we declare images without refs, and that is the responsibility of the docker-app cnab/run tool to actually handle them properly (the implementation does not do file manipulation, but directly modifies in-memory structures after parsing).

technosophos:
Okay, I have come around to @simonferquel and @glyn 's point of view. I now think that the `ref` field should be dropped.

The determining factor to me was realizing that keeping `refs` as spec'ed either requires (a) the invocation images to get very complex, or (b) the CNAB Runtime to modify the invocation image. Both have some deep problems.

As @simonferquel commented a few days ago, the "general purpose" `refs` field is actually more complex than just delegating the substitution task directly to the authors of the invocation image. (@glyn also made a similar point).

So I am going to redo this section to simply remove the `refs` section from the spec.

101-bundle-json.md
technosophos: @glyn you make some very good points there.

For now, I am going to focus on removing the `refs` section as part of this PR. Would you mind opening a new issue that specifically deals with adding whatever is necessary to make it the bundle's responsibility? We can then shift the related conversation to that thread and not lose track of it.

glyn: > @glyn you make some very good points there.
> 
> For now, I am going to focus on removing the `refs` section as part of this PR.

Thanks!

> Would you mind opening a new issue that specifically deals with adding whatever is necessary to make it the bundle's responsibility? We can then shift the related conversation to that thread and not lose track of it.

https://github.com/deislabs/cnab-spec/issues/122

technosophos:
I am now closing this issue, superseded by #122 and #123

PR #121, opened 2019-02-26
docs: Move claims to separate specification

### LINKED ISSUE #84: Should claims really be part of the spec ?
It is more a question than a suggestion.

IMO, claims seem more to be an implementation detail of Duffle. It can be used as a simple way to describe how a CNAB implementation can keep track of existing installations, as well as what data must be stored for future invocations as part of the runtime specification.

technosophos:
This removes 104 and adds 400, with largely the same content. A few clarifications were made along the way in order to move all mention of claims from CNAB Core. Schema and example have been added.

Closes #84

### DIFF
--- a/103-bundle-runtime.md
+++ b/103-bundle-runtime.md
@@ -45,7 +45,7 @@ The _action_ is the action name. It MUST be either one of the built-in actions o
 
 A `CNAB_REVISION` SHOULD be passed into an install operation, and MUST be passed into `upgrade` and `uninstall`, where this is a _unique string_ indicating the current "version" of the _installation_. For example, if the `my_installation` installation is upgraded twice (changing only the parameters), three `CNAB_REVISIONS` should be generated (1. install, 2. upgrade, 3. upgrade).
 
-Revisions are regenerated on destructive operations so that one installation may be tracked over various revisions. CNAB Runtimes MUST generate a new `CNAB_REVISION` for every `install`, `upgrade`, or `uninstall` action. That is, if an application is installed once, upgraded twice, and then uninstalled, four revisions should have been generated. For additionally defined targets, a new `CNAB_REVISION` MUST be generated if the target is labeled `"modifies": true`, and MUST NOT be generated if `"modifies": false`.
+Revisions are regenerated on destructive operations so that one installation may be tracked over various revisions. CNAB Runtimes MUST generate a new `CNAB_REVISION` for every `install`, `upgrade`, or `uninstall` action. That is, if an application is installed once, upgraded twice, and then uninstalled, four revisions must be generated. For additionally defined targets, a new `CNAB_REVISION` MUST be generated if the target is labeled `"modifies": true`, and MUST NOT be generated if `"modifies": false`.
 
 A `CNAB_REVISION` SHOULD be a [ULID](https://github.com/ulid/spec).
 
--- a/103-bundle-runtime.md
+++ b/103-bundle-runtime.md
@@ -35,13 +35,23 @@ CNAB_BUNDLE_NAME=helloworld
 CNAB_ACTION=install
 ```
 
-The _installation name_ is the name of the _instance of_ this application. Consider the situation where an application ("wordpress") is installed multiple times into the same cloud. Each installation MUST have a unique installation name, even though they will be installing the same CNAB bundle. Instance names MUST consist of Graph Unicode characters and MAY be user-readable. The Unicode Graphic characters include letters, marks, numbers, punctuation, symbols, and spaces, from categories L, M, N, P, S, Zs.
+The _installation name_ is the name of the _instance of_ this application. The value of `CNAB_INSTALLATION_NAME` MUST be the installation name. Consider the situation where an application ("wordpress") is installed multiple times into the same cloud. Each installation MUST have a unique installation name, even though they will be installing the same CNAB bundle. Instance names MUST consist of Graph Unicode characters and MAY be user-readable. The Unicode Graphic characters include letters, marks, numbers, punctuation, symbols, and spaces, from categories L, M, N, P, S, Zs.
 
-The _bundle name_ is the name of the bundle (as represented in `bundle.json`'s `name` field). The specification of this field is in the [bundle definition](101-bundle-json.md).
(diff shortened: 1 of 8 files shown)


103-bundle-runtime.md
```diff
@@ -35,13 +35,23 @@ CNAB_BUNDLE_NAME=helloworld
...
+### The CNAB Revision Variable
+
+A `CNAB_REVISION` SHOULD be passed into an install operation, and MUST be passed into `upgrade` and `uninstall`, where this is a _unique string_ indicating the current "version" of the _installation_. For example, if the `my_installation` installation is upgraded twice (changing only the parameters), three `CNAB_REVISIONS` should be generated (1. install, 2. upgrade, 3. upgrade).
+
+Revisions are regenerated on destructive operations so that one installation may be tracked over various revisions. CNAB Runtimes MUST generate a new `CNAB_REVISION` for every `install`, `upgrade`, or `uninstall` action. That is, if an application is installed once, upgraded twice, and then uninstalled, four revisions should have been generated. For additionally defined targets, a new `CNAB_REVISION` MUST be generated if the target is labeled `"modifies": true`, and MUST NOT be generated if `"modifies": false`.
```
vdice: Should `four revisions should have been generated` be reworded as `four revisions must be generated` to more appropriately align w/ this detail's `MUST`-iness?

400-claims.md
```diff
@@ -0,0 +1,197 @@
...
+From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/cscope.rb
+```
+
+The CNAB Claims specification does not define where or how records are stored, nor how these records MAY be used by CNAB tooling. However, this specification describes how a CNAB-based system MUST emit the record to an invocation image, and provides some guidance on maintaining integrity of the system.
+
+This is done so that implementors can standardize on a way of relating a release claim (the record of a release) to release operations like `install`, `upgrade`, or `uninstall`. This, in turn, is necessary if CNAB bundles are expected to be executable by different implementations.
+
+### Anatomy of a Claim
+
+While implementors are not REQUIRED to implement claims, this is that standard format for claims-supporting systems.
```
vdice: should `this is that` read `this is the`?

400-claims.md
```diff
@@ -0,0 +1,197 @@
...
+```
+
+- name: The name of the _installation_. This can be automatically generated, though humans may need to interact with it. It MUST be unique within the installation environment, though that constraint MUST be imposed externally. Elsewhere, this field is referenced as the _installation name_.
+- revision: An [ULID](https://github.com/ulid/spec) that MUST change each time the release is modified.
+- bundle: The bundle, as defined in [the Bundle Definition](101-bundle-json.md)
+- created: A timestamp indicating when this release claim was first created. This MUST not be changed after initial creation.
+- updated: A timestamp indicating the last time this release claim was modified
+- result: The outcome of the bundle's last action (e.g. if action is install, this indicates the outcome of the installation.). It is an object with the following fields:
+  - message: A human-readable string that communicates the outcome. Error messages MAY be included in `failure` conditions.
+  - action: Indicates the action that the current bundle is in. Valid actions are:
```
vdice: Should we revisit this list of valid actions (or re-word)?  Thinking of the well-known custom actions added in https://github.com/deislabs/cnab-spec/pull/102.

(If appropriate to amend, but list not introduced/changed by this PR, I would understand relegating as follow-up.)

400-claims.md
```diff
@@ -0,0 +1,197 @@
...
+  - This is accompanied by running the `upgrade` path in the bundle
+- Given an installation's name, mark the claim as deleted.
+  - This is accompanied by running the `uninstall` path in the bundle
+  - XXX: Do we want to allow the implementing system to remove the claim from its database (e.g. helm delete --purge) or remain silent on this matter?
+
+To satisfy these requirements, implementations of a CNAB package manager are expected to be able to store and retrieve state information. However, note that nothing in the CNAB specification tells _how or where_ this state information is to be stored. It is _not a requirement_ to store that state information inside of the invocation image. (In fact, this is discouraged.)
+
+## Injecting Claim Data into an Invocation Image
+
+Complaint CNAB implementations MUST conform to this section.
```
vdice: `s/Complaint/Compliant`

400-claims.md
```diff
@@ -0,0 +1,197 @@
...
+into the invocation container at runtime:
+
+- `$CNAB_INSTALLATION_NAME`: The value of `claim.name`.
+- `$CNAB_BUNDLE_NAME`: The name of the present bundle.
+- `$CNAB_ACTION`: The action to be performed (install, upgrade, ...)
+- `$CNAB_REVISION`: The ULID for the present release revision. (On upgrade, this is the _new_ revision)
+
+> Credential data, which is also injected into the invocation image, is _not_ managed by the claim system. Rules for injecting credentials are found in [the bundle runtime definition](103-bundle-runtime.md).
+
+Parameters that are declared with an `env` key in the `destination` object, their values will be injected as environment variables according to the name specified. Likewise, files will be injected if `path` is set on `destination`.
```
vdice: First sentence may benefit from rewording.  Perhaps something like: `Parameters declared with an env key in the destination object will be injected as environment variables according to the name specified.`

validate.sh
```diff
@@ -7,9 +7,14 @@
 
 set -eou pipefail
 
-for json in $(ls -1 examples/*); do
-  for schema in $(ls -1 schema/*); do
-    echo "Testing json '$json' against schema '$schema'"
-    ajv test -s $schema -d $json --valid
-  done
+for json in $(ls -1 examples/*-bundle.json); do
```
vdice: Perhaps consider adding a comment above each loop describing their purpose?  (Suggests the contributor whose track record of helpful comments has been lacking of late 😅 )

103-bundle-runtime.md

400-claims.md
technosophos: Fixed

400-claims.md
technosophos: Good catch! I totally missed that part, and several of the action names weren't even correct.

400-claims.md
technosophos: A "Complaint CNAB implementation" is one that whines about every single bundle you install. 😆

400-claims.md
technosophos: Oh, man... that sentence is just embarrassing. Sorry about that.

validate.sh
technosophos: You could probably refactor this code to be better than mine. I wanted to be more clever, but gave up and went for simple.

Docs added.

PR #130, opened 2019-03-05
Encourage thick bundle layer deduplication

glyn:
Fixes https://github.com/deislabs/cnab-spec/issues/129

### DIFF
--- a/104-bundle-formats.md
+++ b/104-bundle-formats.md
@@ -39,20 +39,43 @@ Thick bundles MUST contain the bundle descriptor. In addition, a thick bundle MU
 
 A thick bundle SHOULD be encoded as a gzipped TAR. This specification is neutral as to what compression ratio is used.
 
-The internal layout of the TAR should be as follows:
-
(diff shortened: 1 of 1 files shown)


glyn:
Update: relax the restriction that the `artifacts` directory cannot contain subdirectories in addition to image TAR files. This allows hybrid bundles containing an OCI image layout _and_ other types of images.

glyn:
> +1 I am in favor of of removing the no subdirectory constraint in this manner. If I am understanding this correctly, we'll see the most benefit when exporting/importing bundles with support for multiple platform/architecture types by deduplicating layers and the runtime can reconstruct the image(s) to load into local store during something like an `import` by using the information in the `index.json` file.

Yes, that's right. There would also be a benefit for bundles with images that share (large) layers, even if these are all for a single platform/architecture.

simonferquel:
👍 it would actually be very easy to add support for pulling thick bundles of this kind in https://github.com/docker/cnab-to-oci (as this is pretty close to the way we represent bundles in OCI indexes)

PR #133, opened 2019-03-06
Introduce CNAB Outputs to the Spec

### LINKED ISSUE #132: Add Outputs to CNAB Core Spec
## Problem Statement ##
When an invocation image is run, the bundle author may wish to expose some data. This could be something like a generated hostname, port number or some credential like a Kubernetes config file that the person running the bundle will need to use later on. Currently, we don’t specify a way to handle these in a way that makes it really portable from runtime to runtime.

There should be a mechanism for bundle authors to specify if there ARE outputs from the bundle, and what they are if they do exist, just like we do for inputs (credentials and parameters). 

## Impact ##

Adding the concept of outputs enables two use cases:

1. As a user, I used CNAB to create something and that process results in things that are necessary to access it (but can’t be provided as parameters or credentials) and I want to access the thing
1. As a tooling author, I want to provide a way for bundle authors to easily compose existing CNAB bundles by wiring outputs from one into another

jeremyrickard:
When an invocation image is run, the bundle author may wish to expose some data. This could be something like a generated hostname, port number or some credential like a Kubernetes config file that the person running the bundle will need to use later on. Currently, we don’t specify a way to handle these in a way that makes it really portable from runtime to runtime. This PR adds language specifying how these can be handled at the bundle level, but doesn't specify how tooling has to handle these (i.e. do they get persisted locally, do they get stuck in storage somewhere, are they just dumped to std out?)

Closes #132


jeremyrickard:
Need to update the schemas, but I thought I'd open this now and get feedback before doing that hard(er) part :)

simonferquel:
I really like it, but I have a few issues in mind:
- should we add a section about how a runtime implementation will surface these output values to the user ?
- if the invocation image produces sensitive data (private keys, api-keys etc.) we might don't want to persist this in clear format on disk. Maybe we need some encryption mechanism, or use some kind of in-memory volatile volume for storing these outputs ? What about non-docker drivers ? 
- I wonder if in a v2 of the standard, an invocation image might in fact implement a grpc service so that we could avoid passing secrets, output values etc. using the container RootFS.

jeremyrickard:
@simonferquel I originally thought maybe we shouldn't specify how it gets surfaced to allow the tooling to do whatever the tooling author thought was best. I'm open to amending this to add some specificity there. 

Re: 
_if the invocation image produces sensitive data (private keys, api-keys etc.) we might don't want to persist this in clear format on disk. Maybe we need some encryption mechanism, or use some kind of in-memory volatile volume for storing these outputs ?_ 

That's a good question. That sort of applies to credentials as well no? Files (i.e. probably a volume?) seemed to be the common ground between Docker based drivers and later VM based drivers.

101-bundle-json.md
```diff
@@ -573,5 +602,35 @@ The fields are defined as follows:
...
+        "path" : "/cnab/app/outputs/port"
+    },
+    "clientCert" : {
+        "type" : "file",
+        "path" : "/cnab/app/outputs/clientCert"
+    }
+}
+```
+
+- `output`: name/value pairs describing an application output
```
radu-matei: Should this be `outputs`?
Also, indicating it’s a map might be useful, maybe?

101-bundle-json.md
```diff
@@ -573,5 +602,35 @@ The fields are defined as follows:
...
+  - `<name>`: The name of the output. This is REQUIRED. In the example above: `hostName`, `port`, `clientCert`. This
+    is mapped to a value definition, which contains the following fields:
+    - `type`: "null", "boolean", "string", "number", or "integer" or "file". These types correspond to the main primitive types of JSON (with _integer_ and _file_ as a special case). (REQUIRED)
```
radu-matei: Perhaps link / more detailed explanation of the distinction between “integer” and “number”?

silvin-lubecki:
I guess the schemas should be updated too.

jeremyrickard:
OK, I've updated the schema, examples and the 101-bundle-json.md file to reflect the comments above. I think this is ready for an actual review @simonferquel @silvin-lubecki @radu-matei @technosophos

itaysk:
perhaps the outputs should be represented somehow in the claim->result object?

101-bundle-json.md
```diff
@@ -65,6 +66,22 @@ The following is an example of a `bundle.json` for a bundled distributed as a _t
     }
   ],
   "name": "helloworld",
+  "outputs": {
+    "clientCert" : {
+      "path" : "/cnab/app/outputs/clientCert",
+      "type" : "file"
+    },
+    "hostName" : {
+        "apply-to" : ["isntall"],
```
silvin-lubecki: typo: install

101-bundle-json.md
```diff
@@ -83,76 +100,94 @@ The following is an example of a `bundle.json` for a bundled distributed as a _t
...
       },
-      "minValue": 10,
-      "type": "int"
-    }
-  },
-  "schemaVersion": "v1.0.0-WD",
-  "version": "1.0.0"
-}
+      "hostName" : {
+          "apply-to" : ["isntall"],
```
silvin-lubecki: typo: install

jeremyrickard:
Had a chat offline with @simongdavies about this PR. He suggested that we probably need the ability to flag outputs (and possibly parameters) as sensitive (or secret) so that runtimes can properly handle things that may not be proper to log/output (passwords). Most recent commit adds this.

jeremyrickard:
@youreddy @ryanmoran do you think it makes sense to leverage JSON schema here to define outputs since parameters are now based on JSON Schema?

ryanmoran:
@jeremyrickard It makes sense to me that `outputs` would also adopt JSON Schema. I imagine that a runtime implementation connecting bundles via this feature would mostly mean piping some of the `outputs` of one bundle into the `parameters` of another. In that case, ensuring that the `outputs` line up structurally with the `parameters` would seem important.

jeremyrickard:
@ryanmoran ok thanks for the sanity check on that. I'll update this. 

The use case of a runtime wiring two bundles together is exactly one of the things we (@carolynvs  and I) wanted to solve by introducing this (aside from the  obvious one of "i just created someone, now I need something it created"). Do ya'll have a need for that sort of dependency or chaining as well? It might be good to chat about how we see that working (do runtimes just optionally support it, or do we need to codify something in the spec about dependencies between bundles) .

ryanmoran:
@jeremyrickard We definitely plan on supporting inter-bundle dependencies in our runtime implementation. This `outputs` proposal gets us most of the way to what we would want, but we have some further concerns that I've outlined in #156.

101-bundle-json.md
```diff
@@ -138,9 +157,27 @@ And here is how a "thick" bundle looks. Notice how the `invocationImage` and `im
...
   ],
   "name": "helloworld",
+  "outputs": {
+    "clientCert" : {
+      "path" : "/cnab/app/outputs/clientCert",
+      "sensitive" : true,
+      "type" : "file"
+    },
+    "hostName" : {
+        "apply-to" : ["isntall"],
```
michelleN: spelling nit: install

examples/101.02-bundle.json
```diff
@@ -39,6 +39,23 @@
...
   ],
   "name": "helloworld",
+  "outputs": {
+    "clientCert" : {
+      "path" : "/cnab/app/outputs/clientCert",
+      "sensitive" : true,
+      "type" : "file"
+    },
+    "hostName" : {
+        "apply-to" : ["isntall"],
```
michelleN: spelling nit: install

101-bundle-json.md
```diff
@@ -70,6 +71,23 @@ The following is an example of a `bundle.json` for a bundled distributed as a _t
...
   ],
   "name": "helloworld",
+  "outputs": {
+    "clientCert" : {
+      "path" : "/cnab/app/outputs/clientCert",
+      "sensitive" : true,
+      "type" : "file"
+    },
+    "hostName" : {
+        "apply-to" : ["install"],
```
technosophos: We should really fix `applyTo` here and also in the params section. It's the only one that does not use camel case.

technosophos: Can defer to later change, since it would be a compat breaker

radu-matei:
@jeremyrickard - let me know if you want me to update this branch and merge, I'd be happy to.

jeremyrickard:
@radu-matei @michelleN @silvin-lubecki @ryanmoran Sorry for the delay. I didn't take my macbook with me to DockerCon and forgot to push before I left...so here we go.

@technosophos I also corrected the applyTo in the outputs section _only_. I'll do a follow up to get anymore. 

@radu-matei let's give everyone one more day to review and then merge on Wednesday?

101-bundle-json.md
technosophos: @jeremyrickard still `apply-to` here

101-bundle-json.md
```diff
@@ -138,9 +159,29 @@ And here is how a "thick" bundle looks. Notice how the `invocationImage` and `im
...
+  "outputs": {
+    "clientCert" : {
+      "contentEncoding" : "base64",
+      "contentMediaType" : "application/x-x509-user-cert",
+      "path" : "/cnab/app/outputs/clientCert",
+      "sensitive" : true,
+      "type" : "file"
+    },
+    "hostName" : {
+        "apply-to" : ["install"],
```
technosophos: Still `apply-to`

examples/101.02-bundle.json
technosophos: Still `apply-to` and also `isntall`

PR #137, opened 2019-03-12
Include original image names in image map

glyn:
Provide a normative way of specifying the original image name so that
invocation images can map the original to the new image name.

Also tidy up the definition of the image field by using the recommended CAS
SHA form in examples and by deleting the misleading `REGISTRY/NAME:TAG`
syntax.

Fixes https://github.com/deislabs/cnab-spec/issues/134

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -247,18 +247,21 @@ The following illustrates an `images` section:
 
 ```json
 {
-  "images": {
-    "backend": {
-      "description": "backend component image",
-      "digest": "sha256:aca460afa270d4c527981ef9ca4989346c56cf9b20217dcea37df1ece8120685",
-      "image": "example.com/gabrtv/vote-backend:a5ff67...",
-      "imageType": "docker"
-    },
-    "frontend": {
-      "description": "frontend component image",
-      "digest": "sha256:aca460afa270d4c527981ef9ca4989346c56cf9b20217dcea37df1ece8120685",
-      "image": "example.com/gabrtv/vote-frontend:a5ff67...",
-      "imageType": "docker"
+"images": {
+        "frontend": { 
+            "description": "frontend component image",
+            "imageType": "docker",
+            "image": "example.com/gabrtv/vote-frontend@sha256:aca460afa270d4c527981ef9ca4989346c56cf9b20217dcea37df1ece8120685",
+            "originalImage": "example.com/opendeis/vote-frontend@sha256:aca460afa270d4c527981ef9ca4989346c56cf9b20217dcea37df1ece8120685",
+            "digest": "sha256:aca460afa270d4c527981ef9ca4989346c56cf9b20217dcea37df1ece8120685"
+        },
+        "backend": {
+            "description": "backend component image",
+            "imageType": "docker",
+            "image": "example.com/gabrtv/vote-backend@sha256:bca460afa270d4c527981ef9ca4989346c56cf9b20217dcea37df1ece8120686",
+            "originalImage": "example.com/opendeis/vote-backend@sha256:bca460afa270d4c527981ef9ca4989346c56cf9b20217dcea37df1ece8120686",
+            "digest": "sha256:bca460afa270d4c527981ef9ca4989346c56cf9b20217dcea37df1ece8120686"
+        }
     }
   }
 }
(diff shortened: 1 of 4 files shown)


101-bundle-json.md
```diff
@@ -269,7 +271,8 @@ Fields:
 - `images`: The list of dependent images
   - `description`: The description field provides additional context of the purpose of the image. 
   - `imageType`: The `imageType` field MUST describe the format of the image. The list of formats is open-ended, but any CNAB-compliant system MUST implement `docker` and `oci`. The default is `oci`.
-  - `image`: The REQUIRED `image` field provides a valid reference (REGISTRY/NAME:TAG) for the image. Note that SHOULD be a CAS SHA, not a version tag as in the example above.
+  - `image`: The REQUIRED `image` field provides a valid reference for the image. Note that SHOULD be a CAS SHA, as in the example above, not a version tag.
+  - `originalImage`: The OPTIONAL `originalImage` field provides another valid reference for the image. Note that SHOULD be a CAS SHA, as in the example above, not a version tag. If the `originalImage` field is omitted, its value defaults to that of the `image` field.
```
technosophos: Excellent!

technosophos:
This looks great to me. Just left the comment about regenerating the example.

glyn:
> Would you mind changing the example file that we use to generate 103 as well?
> 
> https://github.com/deislabs/cnab-spec/blob/master/examples/103.1-bundle.json
> 
> Otherwise, we might lose that change next time we copy the example in there.

Done. Also removed the extraneous `@` from the digest.

glyn:
I force pushed an update to change the example file as requested and to make some corrections and improvements.

101-bundle-json.md
simonferquel: Note that the originalImage reference should be grammatically a valid reference, but since it will be replaced at runtime with the "image" value, it can potentially point to a deleted repository.

simonferquel:
Invocation images can be relocated as well (and are in fact relocated when using cnab-to-oci to push an existing bundle to a new repo). Should we introduce `originalImage` on invocation images as well ?

101-bundle-json.md
glyn: > Note that the originalImage reference should be grammatically a valid reference, but since it will be replaced at runtime with the "image" value, it can potentially point to a deleted repository.

Good point! However, I don't really want to encourage people to delete original images. For example, we'd like our users to be able to see that the digest of the original and new image references are identical so that it's certain that the images consist of the same bits. Deleting the original image would stop them being able to obtain its digest.

Also, we currently say that the run tool MAY replace any original image references, so if we want to allow for original images to be deleted, we'd probably want to strengthen that and say the run tool SHOULD replace those references.

glyn:
> Invocation images can be relocated as well (and are in fact relocated when using cnab-to-oci to push an existing bundle to a new repo). Should we introduce `originalImage` on invocation images as well ?

Not sure. That would serve a different purpose as by the time the invocation image is running, who cares what it's original image name was. Maybe best to address this with a separate issue/PR so we can think it through separately.

technosophos:
I am still thinking that if we want support for moving invocation images, we could do it like this:

- Fetch the bundle
- Unpack the invocation images into the local Docker daemon (essentially, Docker Import)
- Execute a specific target designed to handle the mechanics of importing images (A target action like `import`, perhaps?)
- That action's sole responsibility would be to more things into their desired "internal" locations
- To install a bundle, though, one would still need to run a separate `install` action

So the flow of this would be (using Duffle for example, since I'm most familiar with it):

- `duffle export`: packs all the data into a thick bundle
- Move the bundle however you see fit
- `duffle import`: unpacks the data locally, then runs the `import` action on the invocation images
- `duffle install`: runs the regular installation, and assumes that all images are already relocated

The following changes would be required on the spec:

- We'd probably need to standardize an `import` event. We _might_ decide that as output we want it to generate a JSON document describing the two image sections
- We'd need to update the spec to state that doing imports may allow credential and parameter injection (since right now they don't)

simonferquel:
@technosophos I am sorry, but I still disagree with that. Storing bundle and images, and moving them across registries is not something we want implemented in each invocation image. That is a cross-cutting concern that must be implemented in the tooling. The cnab-to-oci effort is a great example of how to do that for images of type `oci` or `docker`: the whole mechanism of publishing in different repositories (or even different registries), and leveraging the current state of the spec to pass the component image references to the invocation image is done by cnab-to-oci. The only thing required in the invocation-image, is to read the image map and use it as the source of truth for knowing which image should be used for the deployed components.

There is also a drawback when using docker import or load: the docker image store do not preserves digested references on import / load.

technosophos:
That is a good point.

glyn:
@technosophos Is there anything that needs to be done before this PR can be merged? I think it would be reasonable to treat relocation of invocation images separately, if necessary.

103-bundle-runtime.md
```diff
@@ -234,10 +235,11 @@ The `/cnab/app/image-map.json` file mounted in the invocation image will be:
 {
     "my-microservice": {
         "description": "my microservice",
-        "digest": "sha256:aaaaaaaaaaaa...",
-        "image": "technosophos/microservice:1.2.3"
+        "image": "technosophos/microservice@sha256:cca460afa270d4c527981ef9ca4989346c56cf9b20217dcea37df1ece8120687",
+        "originalImage": "gabrtv/microservice@sha256:cca460afa270d4c527981ef9ca4989346c56cf9b20217dcea37df1ece8120687",
+        "digest": "sha256:cca460afa270d4c527981ef9ca4989346c56cf9b20217dcea37df1ece8120687"
```
michelleN: I'm a little thrown off by the digest field. Does it relate to `originalImage` or `image` here?

glyn: Well spotted! I was assuming that the digest is preserved, as I regard that as a pretty important property of image relocation. If we really want to accommodate non digest preserving relocation, perhaps we need an `originalDigest` field as well.

michelleN: that sounds like a good solution to me!

technosophos:
Do we still need an originalDigest field, or can this be merged as-is?

glyn:
Added the `originalDigest` field.

103-bundle-runtime.md
glyn: Added.

glyn:
See also https://github.com/deislabs/cnab-spec/pull/158 which adds `originalImage` and `originalDigest` fields to invocation images.

simonferquel:
It seems really weird to me that the digest might change on relocation. If it the case, it implies that the digest is incorrectly generated. It feels like this digest should be used for validating that the image is still the same, even if it has been relocated.

103-bundle-runtime.md
simonferquel: it does not makes sense IMO. The digest should not change depending on the image location

michelleN:
@simonferquel It really depends on which digest you're talking about which we never clarify in the spec and has been a source of confusion. The question is how do you build an image and attain a digest that is unrelated to the name/tag of the image. Glyn sent me some possible sources for help that I'm going to look into but I think we need to open a separate issue for this topic.

simonferquel:
I played a bit with containerd code base, and actually, there is a pretty good way to do so.
For a bit of context, digesting the archive output from `docker save` is not a good idea, because it does not only contain the image manifest/config/blobs, but also a layout mapping repo/tags to image configs.

If you look at https://github.com/containerd/containerd/blob/master/cmd/ctr/commands/images/import.go, it is able to load an archive, and populate separately a content store (which is content addressable - like a local registry), and a label store. The digest it returns does not depend on the image repo/tag, and thus can be used for stable digests.
If you are only interested in the digest, you can implement a discarding content.Store, if you want to use this for implementing bundle export, you can just use the local store implementation from containerd.

simonferquel:
Hmm sorry, correct file in containerd is: https://github.com/containerd/containerd/blob/master/images/archive/importer.go

michelleN:
That's wonderful @simonferquel. Thanks for pointing me to that. I didn't realize this was or could be the case until @glyn pointed it out yesterday and now you are pointing out the exact implementation I've been looking for.

glyn:
For the record, https://github.com/pivotal/image-relocation preserves digests during relocation courtesy of the design of its https://github.com/google/go-containerregistry dependency.

PR #139, opened 2019-03-14
Various minor fixes

radu-matei:
Re-reading parts of the spec resulted in a few little nits.

### DIFF
--- a/100-CNAB.md
+++ b/100-CNAB.md
@@ -20,18 +20,18 @@ The _bundle definition_ is a single file that contains the following information
 - The list of executable images that this bundle will install
 - A list of credential paths or environment variables that this bundle requires to execute
 
-The canonical encoding of a bundle definition is a JSON-formatted file, which MUST be encoded as a Canonical JSON Object stored in a `bundle.json` file, as defined in [the bundle file definition](101-bundle-json.md)
-
-In either case, CNAB has the same schema, and this spec refers to this file as the "bundle definition" (or occasionally "bundle file"). 
+The canonical encoding of a bundle definition is a JSON-formatted file, which MUST be encoded as a Canonical JSON Object stored in a `bundle.json` file, as defined in [the bundle file definition](101-bundle-json.md).
 
 However, as a signed bundle definition represents an immutable bundle, all invocation images and images references must have a digest.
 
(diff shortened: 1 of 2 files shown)


100-CNAB.md
```diff
@@ -20,18 +20,18 @@ The _bundle definition_ is a single file that contains the following information
...
+The bundle definition can be stored on its own, or as part of a _packaged archive_, which is a CNAB bundle that includes the JSON file and exported images (including the [invocation image](102-invocation-image.md)).
 
 - A _thin bundle_ consists of just a bundle definition.
 - A _thick bundle_ consists of a packaged archive that contains both the bundle definition and an encoded representation of all of the invocation images and images.
 
-When thin bundles are processed, the referenced content (such as invocation images and other images) are retrieved from their respective storage repositories and registries. A bundle is considered to be _well formed_ if it's definition follows the CNAB schema and the images are in the correct formats. A bundle is considered _complete_ if it is packaged as a thick bundle, and all the components are present OR if it is packaged as a thin bundle and all of the references are resolvable. Completeness is thus in some cases contingent upon external factors such as network access.
+In either case, CNAB has the same schema, and this spec refers to this file as the "bundle definition" (or occasionally "bundle file").
```
radu-matei: This sentence was moved because I think "either case" refers to think vs. thin, and it makes sense here.

simonferquel: About Thick bundles, I think we actually are missing a very important thing: the CNAB runtime can access component images, but we gave no way for the invocation image to understand that it is invoked from an "off-line" bundle, and how it can access those images.

If you take the example of a generic "compose" CNAB bundle, if you want it to work offline with a thick bundle, the invocation image must:
- know that it is invoked in "off-line" mode (meaning it must first somehow provision component images on the cluster, instead of creating workloads referencing images from a registry)
- has a way to access the images within the thick bundle.

@technosophos wdyt about that, is there something already designed for air-gaped installations with component images ?

I am sorry it comes a little late in the process, but I think if we want air-gaped from version 1, we need to tackle this subject.

technosophos: /cc @michelleN

technosophos:
Okay, I think we should merge this one as-is. I think @simonferquel 's point is related to several other issues with thick bundles that are being treated differently, but the current text of this PR is good.

That okay with you Simon?

PR #145, opened 2019-03-20
Add `schema` field to parameters section

youreddy:
This PR allows bundle authors to specify additional parameter types (such as arrays and objects) through a new field: `schema`. The referenced schema document must conform to the JSON schema specification.

This PR is a follow up from the conversation started [in a proposal to introduce complex objects](https://github.com/deislabs/cnab-spec/issues/114#issuecomment-467182356) and addresses the first action item listed [here](https://github.com/deislabs/cnab-spec/issues/114#issuecomment-474096981). 

Instead of specifying `schema` as an inlined JSON Schema document, we decided to allow bundle authors to embed a schema document within the bundle. This is mainly because of the user experience issues described in this [slack thread](https://cloud-native.slack.com/archives/CEX1W7WMD/p1553011275226300).

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -308,6 +308,7 @@ Parameter specifications are flat (not tree-like), consisting of name/value pair
   - `<name>`: The name of the parameter. This is REQUIRED. In the example above, this is `backend_port`. This
     is mapped to a value definition, which contains the following fields:
     - `type`: "null", "boolean", "string", "number", "integer", "bytes". These types correspond to the main primitive types of JSON (with _integer_ and _bytes_ as special cases). Objects and arrays are not supported, as the serialization of these is underdetermined. (REQUIRED)
+    - `schema`: A reference to a JSON schema document that is used to validate the structure of the parameter. Only applies to string parameter types (OPTIONAL)
     - `required`: if this is set to true, a value MUST be specified (OPTIONAL, not shown)
     - `defaultValue`: The default value. For type `bytes`, defaultValue is a string containing a base64 payload (OPTIONAL)
     - `enum`: an array of allowed values. For type `bytes`, values must be strings containing a base64 payload (OPTIONAL)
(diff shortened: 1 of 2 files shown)


101-bundle-json.md
```diff
@@ -333,6 +334,62 @@ Some validation terms have aliased keywords for backward compatibility with ARM
...
+            "path": "<fully-qualified-path>"
+        },
+        "apply-to": ["action1", "action2"]
+    }
+}
+```
+
+See [The Bundle Runtime](103-bundle-runtime.md) for details of how parameters are injected into the invocation image.
+
+### Specifying Schema References
```
technosophos: Thoughts on adding schema support to credentials, too? Azure credential files, for example, are schematized JSON. (Definitely the kind of thing we could defer to a follow-up PR)

101-bundle-json.md
```diff
@@ -333,6 +334,62 @@ Some validation terms have aliased keywords for backward compatibility with ARM
...
+### Format of Parameter Specification
+
+The structure of a parameters section looks like this:
+
+```json
+"parameters": {
+    "<parameter-name>" : {
+        "type" : "<type-of-parameter-value>",
+        "schema": "<reference-to-schema-document>",
+        "required": true|false
```
technosophos: minor nit: missing trailing `,`

technosophos:
Okay, so here's my view on this:

1. For background, the sorta JSON Schema-ish notation that is currently there was originally intended to _feel_ like JSON Schema (as opposed to the ARM Template syntax we started out with). But I hadn't thought of just _using_ JSON Schema. That's why I wasn't particularly picky about things like `bytes`
2. If we are to support arbitrarily complex configuration options, I am definitely okay with using JSON + JSON Schema as the way to do this. It feels like the right fit.
3. The crux of the argument seems to be these two opposing views:
  - For: Since we can validate simple types, we should also provide a way of supporting complex types with the same sort of validation. JSON/JSON Schema seems like the sensible way to do this
  - Opposed: Validation of complex types ought to be left to the invocation image runtime, and the simple validation rules are enough to get what we need today
4. The technical implementation outlined above looks great to me. I don't have any qualms with them. (Though we probably will have to figure out a way to inject external schemata for the export/import story)

My view on (3) is that it makes sense to support validation of complex objects in the CNAB Runtime, and my argument in favor of this is based on a way to construct a user experience: I would like to be able to build a UI generator that could dynamically build a user interface (like a web form or something), and then give the user immediate feedback on the validity of their parameters before beginning the installation process. With a JSON Schema, that process could be pushed all the way into the browser with very little effort.

I agree with Simon that this doesn't _replace_ the need for doing validation inside of invocation images. Nor can it necessarily perform _all_ of the necessary validations. However, I think this is an acceptable trade-off.

youreddy:
@technosophos to clarify, are you in favor of having the entirety of parameters being JSON Schema? Or what this current PR is doing where the `schema` field can reference a definition that is JSON schema conformant? 

Our preference would be to have parameters just be JSON Schema and we could replace this PR with one that does just that. In that case, supporting `bytes` could be achieved through having an optional [format](https://json-schema.org/latest/json-schema-validation.html#format) key.

squillace:
@youreddy per the conversation in the public stand up, I LIKE the idea of json-schema quite a bit over "this and that" approach. I would like to think about what the spec should say or not say about _where_ the validation of schema might occur if the element exists. The proposed second PR would help everyone think on it. This is a good issue.

youreddy:
In the [CNAB community meeting](https://hackmd.io/s/SyGcBcwQ4#March-27-Agenda) today, we determined that there's enough interest in having the `bundle.json` be fully JSON schema conformant. The PR above would have perpetuated the pattern of having the parameters section be subtly different from JSON schema. We are going to close this PR in favor of a new one that has a JSON schema version of `bundle.json`.

There's an open question about on whether any client tooling (such as the VS Code extension) should be required to validate if a bundle is conformant to the JSON schema specified. The current thinking is that clients could optionally perform this validation but we will have a better understanding once the new PR is opened. 

/cc @michelleN @radu-matei @squillace

youreddy:
New PR: https://github.com/deislabs/cnab-spec/pull/153

PR #146, opened 2019-03-20
add front matter for rendering a hugo website 

flynnduism:
This PR adds a small bit of metadata to each page, in order to structure the markdown content for rendering as a website:

The site is a [work in progress](https://github.com/deislabs/cnab.io/pull/10) - [github.com/deislabs/cnab.io/pull/10](https://github.com/deislabs/cnab.io/pull/10)

It will fetch the contents of this repo and import it into cnab.io - rather than adding Hugo _config/theme/noise_ to this repo.

### DIFF
--- a/804-repositories.md
+++ b/804-repositories.md
@@ -1,6 +1,5 @@
 ---
-title: CNAB Registries
-weight: 804
+title: Registries
 ---
 
 # CNAB Registries
--- a/804-repositories.md
+++ b/804-repositories.md
@@ -1,3 +1,8 @@
+---
+title: CNAB Registries
+weight: 804
+---
+
 # CNAB Registries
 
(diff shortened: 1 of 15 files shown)


804-repositories.md
```diff
@@ -1,3 +1,7 @@
+---
+title: Registries
```
technosophos: Is registries on a `weight` loss program?

805-well-known-custom-actions.md
```diff
@@ -1,3 +1,7 @@
+---
+title: Well known custom actions
```
technosophos: Also seems to have lost some `weight`

804-repositories.md
flynnduism: the file `804-` is redundant, as the content is now in file `200-` so I've purposely left it out of the website indexing here.

805-well-known-custom-actions.md
flynnduism: Thanks for catching! Updated.

PR #153, opened 2019-03-28
Converts parameters section to extend JSON Schema

ryanmoran:
As requested during the March 27 community meeting, here is a new PR that outlines the proposal to change the `parameters` section to extend JSON Schema. The primary driver behind this PR is to provide more complex configuration inputs like arrays and object with significantly improved validation of those inputs. Additionally, this change serves to reduce confusion between the "JSON Schema"-alike `parameters` section of the current spec and JSON Schema.

This PR still leaves unresolved the concern of the representation of these complex parameter types inside the invocation image. If we support complex configuration then we will need the spec to make a determination on the format of that data when it is placed at a `destination`.

/cc @youreddy

### DIFF
--- a/schema/bundle.schema.json
+++ b/schema/bundle.schema.json
@@ -105,11 +105,16 @@
         "parameters":{
             "description": "Parameters that can be injected into the invocation image",
             "type": "object",
+            "properties": {
+                "required": { "$ref": "http://json-schema.org/draft-07/schema#/properties/required" }
+            },
             "additionalProperties": {
-                "$ref": "#/definitions/parameter"
+                "allOf": [
+                    { "$ref":  "#/definitions/parameter" },
+                    { "$ref":  "http://json-schema.org/draft-07/schema#" }
+                ]
             }
         }
-        
     },
     "required": [ "name", "version", "invocationImages", "schemaVersion"],
     "definitions": {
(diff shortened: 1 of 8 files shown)


schema/bundle.schema.json
```diff
@@ -104,12 +104,16 @@
         },
         "parameters":{
             "description": "Parameters that can be injected into the invocation image",
-            "type": "object",
```
michelleN: was `"type": "object"` purposefully deleted?

ryanmoran: Nope. That is a mistake. We will revert that bit.

simonferquel:
I like the fact that we align with the format with JSON schema formalism. The only thing we must be really clear about is that it does not mean we consider the input values to be json by themselved: `the value` is a perfectly valid string from a CNAB point of view, and we don't want to force user to wrap that in a JSON-valid format (such as `"the value"`).

simonferquel:
A duffle command might be easier to spot the difference:
`duffle install <...> -p myparam="some text"` -> currently works but the for the runtime myparam value is not quoted. to make it work in a pure json world, it would have to be:
`duffle install <...> -p myparm="\"some text\""`. This something we don't want (also because, every invocation image would have to json-decode any parameter value as well.

michelleN:
@simonferquel - that seems like a runtime concern and we'll definitely handle that experience on our end. Given that, do you have any other objections or can we move forward with this addition?

michelleN:
I'm going to merge given our consensus on the call but @simonferquel please feel free to open up any follow up issues.

PR #160, opened 2019-04-11
rename digest to content-digest for clarity

### LINKED ISSUE #159: rename digest field to contentDigest

michelleN:
as suggested by @simonferquel 
resolves #159


103-bundle-runtime.md
```diff
@@ -195,13 +195,13 @@ For this example CNAB bundle:
...
       "description": "my microservice",
-      "digest": "sha256:aaaaaaaaaaaa...",
+      "content-digest": "sha256:aaaaaaaaaaaa...",
       "image": "technosophos/microservice:1.2.3"
     }
   },
   "invocationImages": [
     {
-      "digest": "sha256:aaaaaaa...",
+      "content-digest": "sha256:aaaaaaa...",
```
radu-matei: I think there is one more reference of `digest` in this file, in the example below (line 240): "The `/cnab/app/image-map.json` file mounted in the invocation image will be"

michelleN: good catch. updated.

100-CNAB.md
```diff
@@ -29,7 +29,7 @@ The canonical encoding of a bundle definition is a JSON-formatted file, which MU
 
 In either case, CNAB has the same schema, and this spec refers to this file as the "bundle definition" (or occasionally "bundle file"). 
 
-However, as a signed bundle definition represents an immutable bundle, all invocation images and images references must have a digest.
+However, as a signed bundle definition represents an immutable bundle, all invocation images and images references must have a content-digest.
```
simonferquel: or have a digested image reference (e.g.: myimage@sha256:aabb...)

simonferquel:
I am not sure about the correct wording for my comment above, but I think we really need some clarification here.

radu-matei:
@michelleN, @simonferquel - should I add this topic for this week's CNAB meeting?
I know @michelleN probably won't be able to make it to the meeting, but I think we can at least clarify what we're after here.

schema/bundle.schema.json
```diff
@@ -131,7 +131,7 @@
                     "type": "string",
                     "default": "oci"
                 },
-                "digest":{
+                "content-digest":{
```
jlegrone: Seems like there's a strong convention around camel cased keys.

100-CNAB.md
michelleN: I disagree here. At the point at which a bundle is ready to be shared or shipped, a bundle should have a content digest regardless of whether the image reference contains a digest or not. Do you disagree?

michelleN:
@simonferquel Since this comment was from a long while ago, do you still feel thin bundles should not have content digest. I think both thin/thick bundles rather any bundle that is ready to be shared or shipped should contain content digests for each image it references. What are your current thoughts around this?

>It would be nice to clarify the content-digest conditional requirements:

>>if in-dev: no digest required, but bundle is not shareable
>>if thick-bundle: content-digest is required
>>if thin-bundle: content-digest OR digested reference is required

michelleN:
@technosophos @jlegrone updated. thanks

simonferquel:
My point was more about "in-dev" bundles: basically a bundle is not ready to be shared until everything is either referenced by digest or has a contentDigest with it. But for making things easier during development, the runtime should be capable of accepting not-digested reference (e.g. images I just built locally). The spec does not have to specify how the tooling should enforce this though.

simonferquel:
otherwise, LGTM

technosophos:
@michelleN This needs a rebase, but I think you can merge after that (unless you wanted to address Simon's last comment)

michelleN:
@simonferquel - I added some wording that addresses your concerns. Lmk what you think!

Issue #156, opened 2019-04-10
How can runtimes help bundle authors with dependencies?


ryanmoran:
There is already an outstanding [issue](https://github.com/deislabs/cnab-spec/issues/132) and [PR](https://github.com/deislabs/cnab-spec/pull/133) that enables bundles to share configuration data. In addition to this configuration concern, I think there is also a question of how bundles might declare further runtime dependencies.

As an example, imagine that I am building an application bundle that depends upon a separate database bundle. The current thinking around `outputs` would cover much of my current concerns for how my database bundle might share configuration with my application bundle.

#### Database Bundle `outputs`
```json
{
  "outputs": {
    "address": {
      "type": "string",
      "path": "/cnab/app/outputs/address"
    },
    "port": {
      "type": "integer",
      "path": "/cnab/app/outputs/port"
    },
    "name": {
      "type": "string",
      "path": "/cnab/app/outputs/name"
    },
    "username": {
      "type": "string",
      "path": "/cnab/app/outputs/username"
    },
    "password": {
      "type": "string",
      "path": "/cnab/app/outputs/password"
    }
  }
}
```

#### Application Bundle `parameters`
```json
{
  "parameters": {
    "address": {
      "type": "string"
    },
    "port": {
      "type": "integer"
    },
    "name": {
      "type": "string"
    },
    "username": {
      "type": "string"
    },
    "password": {
      "type": "string"
    }
  }
}
```

But this implementation doesn't really enforce a particular dependency of my application, which is that it requires a PostgreSQL database, and will break horribly if I accidentally configure it with MySQL.

Maybe I could add an `adapter` parameter to my bundle like the following:
```json
{
  "parameters": {
    "address": {
      "type": "string"
    },
    "port": {
      "type": "integer"
    },
    "name": {
      "type": "string"
    },
    "username": {
      "type": "string"
    },
    "password": {
      "type": "string"
    },
    "adapter": {
      "type": "string",
      "const": "postgres"
    }
  }
}
```

I can then enforce this dependency if the database bundle author then includes an output like the following:
```json
{
  "outputs": {
    "address": {
      "type": "string",
      "path": "/cnab/app/outputs/address"
    },
    "port": {
      "type": "integer",
      "path": "/cnab/app/outputs/port"
    },
    "name": {
      "type": "string",
      "path": "/cnab/app/outputs/name"
    },
    "username": {
      "type": "string",
      "path": "/cnab/app/outputs/username"
    },
    "password": {
[...]

Issue #65, opened 2018-12-21
Dependency Management Model for bundles


nadundesilva:
I've been going through the Spec and I was thinking about how a user can create a bundle that depends on multiple bundles. In the current Spec, we do have the ability to share bundles and thereby reuse them. But there is no concept of dependencies between them.

In a distributed system composed of many components running on heterogeneous environments, we would probably create multiple bundles for different components and deploy them. If I am to use a bundle that depends on other bundles, I need to separately pull each bundle and then install them separately.

However, when the number dependencies gets larger, finding all the dependencies and manually pulling them would be really hard for a user.

Do we have any plans to address this issue through the Spec?

In my opinion, we would have to introduce a Dependency Management Model for the bundles. By embedding this into the Spec at an early stage we can make sure that the bundles can be scaled well into larger distributed systems with many components.

Unfortunately, I do not have an exact proposal for this as of now. However, in my opinion we would have to consider the following if we are going to introduce this. (I am willing to contribute, if this seems like something you would like to add to the Spec)

1. Specifying the dependencies in the bundle.json file
2. Specifying how the dependencies would be shared (we can probably adjust the repository section in the Spec to support this.)
3. Specify how pulling from the repository should resolve the dependencies and install them.
4. What should be the policy when multiple bundles require the same bundle as a dependency

WDYT?

nadundesilva:
After taking a look at [Porter](https://github.com/deislabs/porter), it looks like you have already planned to introduce a Dependency Model for CNAB. Do you have any plans to add it to the Spec as well?

carolynvs:
Porter works around the lack of dependencies in the runtime spec by identifying dependencies before the invocation image is built, and then including them in the invocation image at build time. The final result is a single invocation image and the tool executing the bundle (duffle) is unaware that dependencies were involved at all.

garethr:
This is similar to the view we came to thinking about this in docker app as well. That dependencies are more of a build-time concern in CNAB. I'll try and rationalise that view.

> In a distributed system composed of many components running on heterogeneous environments, we would probably create multiple bundles for different components and deploy them.

I think this is definitely true. If we take a whole system view that's not one application, but multiple. That doesn't say anything about dependencies however.

> I need to separately pull each bundle and then install them separately.

[...]

PR #166, opened 2019-04-25
Dependencies Spec: Named Dependencies

carolynvs:
This is the first step towards #156. I decided to submit a proposal for specifying named dependencies as a first step, instead of jumping directly to what I proposed in the comments of that issue. I explain a bit more in the Q&A (see md file). Hopefully this is merged and then we can move onto that! 😇 

It's also related to #65 which honestly is a HUGE issue and this only covers a tiny bit.

cc @ryanmoran

### DIFF
--- a/500-CNAB-dependencies.md
+++ b/500-CNAB-dependencies.md
@@ -66,7 +66,7 @@ The `requires` object defines the criteria for the dependent bundle:
 * `bundle`: A reference to a bundle in the format REGISTRY/NAME.
 * `version`: A set of criteria applied to the bundle version when selecting an
     acceptable version of the bundle.
-    * `range`: An allowed version range. 
+    * `ranges`: An array of allowed version ranges. 
     
         Versions are specified using semver, with or without a leading "v". 
         An `x` in the place of the minor or patch place can be used to specify 
--- a/500-CNAB-dependencies.md
+++ b/500-CNAB-dependencies.md
@@ -1,0 +1,83 @@
+---
+title: Dependencies
+weight: 500
+---
+
+# CNAB Dependencies 1.0
+Working Draft, April. 2019
+
+This specification describes how bundles can define dependencies on other
+bundles.
+
+This specification, the CNAB Dependencies specification, is not part of the CNAB Core specification. An implementation of CNAB Core MAY NOT implement this specification, yet still claim compliance with CNAB Core. A CNAB implementation MUST NOT claim to be CNAB Dependencies-compliant unless it meets this specification.
+
+# Overview
+
+One way for a bundle author to manage the complexity of a large bundle and reuse
+common logic is to create bundles for discrete components, such as managing
+a database as a service on a cloud platform, and then use that bundle as a 
+dependency.
+
+This specification defines [dependencies metadata](#dependencies-metadata) in the 
+bundle.json for specifying dependencies but does not dictate that the metadata is
+specifically used at a particular time, or how the dependency graph is resolved. 
+Each tool may choose to use the information differently, for example [porter](https://porter.sh) resolves dependencies when the bundle is built, but another tool could 
+use this information at runtime.
+
+There are two cases for how a bundle may need to depend upon another bundle:
+
+1. [Depend on a named bundle](#depend-on-a-named-bundle)
+1. [Depend on a bundle that provides a capability](#depend-on-a-bundle-that-provides-a-capability)
+
+## Depend on a named bundle
+
+The bundle depends on a specific named bundle that is known in advance.
+
+```json
+{
+    "name": "wordpress",
+    "dependencies": [
+        {
+            "requires":{
+                "bundle": "azure/mysql",
+                "version": {
+                    "range": "5.7.x",
+                    "prereleases": "true"
+                }
+            }
+        }
+    ]
+}
+```
+
+## Depend on a bundle that provides a capability
+
+This section is a placeholder and will be completed in a follow-up pull request.
+
+## Dependencies Metadata
+
+This specification introduces a `dependencies` array in the bundle.json
+that defines metadata necessary to specify a dependency.
+
+### requires
+
+The `requires` object defines the criteria for the dependent bundle:
+
+* `bundle`: A reference to a bundle in the format REGISTRY/NAME.
+* `version`: A set of criteria applied to the bundle version when selecting an
+    acceptable version of the bundle.
+    * `range`: An allowed version range. 
+    
+        Versions are specified using semver, with or without a leading "v". 
+        An `x` in the place of the minor or patch place can be used to specify 
+        a wildcard. Ranges can be specified by separating the two versions with 
+        a dash, the dash must be surrounded by spaces in order to disambiguate 
+        from prerelease tags.
+
+        Below are some example ranges:
+        * `1.2.3` - Requires an exact version.
+        * `1.2.x` - Restricts the major.minor to `1.2` and allows any patch.
+        * `1.x` - Restricts the major to `1` and allows any minor.
+        * `1.5.x - 3` - Allows a range of versions from `1.5` up to but not including 4.
+    * `prereleases`: A boolean field, defaults to `false`, which specifies if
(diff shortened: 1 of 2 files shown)


ryanmoran:
@carolynvs this looks great as a place to start. Thank you for taking the time to think about this and write it up.

While we eventually want to get to a dependency system that can resolve generalized "capabilities", this is certainly something we could use today. I only have one question about the version ranges. Let me explain some background...

We have an existing set of products that have a similar dependency system. We are able to declare that one product is dependent upon another product within a single version range.

Using this dependency system, we have some software that has a frequent release cadence. Those pieces of software tend to maintain long-term support over some N minor releases. So, if I have a product at version 4.6, then I am likely maintaining LTS releases for 4.5, 4.4, and 4.3.

Also using this dependency system are some products that have a more infrequent release cycle. Those products might be dependent on a product with a set of LTS releases, and would be expected to maintain compatibility.

As an example, suppose that I have a "frequent release" product called `FAST` and I have an "infrequent release" product called `SLOW`. The `SLOW` product declares a dependency across a number of minor releases of the `FAST` product.
```json
{
    "name": "SLOW",
    "dependencies": [
        {
            "requires":{
                "bundle": "FAST",
                "version": {
                    "range": "4.3.x - 4.6.x",
                    "prereleases": "true"
                }
            }
        }
    ]
}
```

This works great for a long time. Whenever patch releases of `FAST` come out on the 4.3-4.6 minor release lines, my `SLOW` product remains compatible. But, eventually, there is a bug found in 4.5.1 and 4.6.3 of `FAST`. This bug has an impact on `SLOW` and so I want to make sure users only install `SLOW` with patch versions of `FAST` that are fixed. I now have a situation where I want to declare a dependency like this:
```json
{
    "name": "SLOW",
    "dependencies": [
        {
            "requires":{
                "bundle": "FAST",
                "version": {
                    "range": "4.3.x - 4.4 | 4.5.2 - 4.5.x | 4.6.4 - 4.6.x",
                    "prereleases": "true"
                }
            }
        }
    ]
}
```

My dependency range is now non-contiguous ☹️. I could just keep the existing range, but it is unfortunate when users hit an issue caused by the bug found in `FAST` that could have been prevented by having a more restrictive range definition.

Would you consider a more complex version range language to support this type of need?

cc @youreddy

carolynvs:
> Would you consider a more complex version range language to support this type of need?

Yup! I have nothing against having multiple version ranges. I just didn't include it because I forgot about this scenario when I wrote it up. 😊  I believe that the pipes should work well, so I'll update the PR with that.

carolynvs:
I said that then immediately wondered if it wouldn't be easier to work with if we did this instead?

```
"ranges": [ "4.3.x - 4.4", "4.5.2 - 4.5.x" "4.6.4 - 4.6.x" ],
 "prereleases": true
```

I think that would result in less string parsing and is a bit easier to understand when just reading it as a lay person.
"

ryanmoran:
> ranges: [ "4.3.x - 4.4", "4.5.2 - 4.5.x", "4.6.4 - 4.6.x" ]

LGTM. Thanks!

500-CNAB-dependencies.md
```diff
@@ -0,0 +1,100 @@
...
+        * `1.5.x - 3` - Allows a range of versions from `1.5` up to but not including 4.
+    * `prereleases`: A boolean field, defaults to `false`, which specifies if
+        prerelease versions of the bundle are allowed.
+
+# Q & A
+
+This isn't meant to be in the final spec but gives people a place to comment
+on anticipated questions.
+
+* Why didn't use you `~>` and other cute symbols?
```
radu-matei: Thank you!

carolynvs:
This is ready for review now! 😀

500-CNAB-dependencies.md
```diff
@@ -0,0 +1,100 @@
...
+    * Because they have no global meaning across languages and frameworks and are hard to understand at a glance.
```
technosophos: I don't find that answer convincing: (a) The meaning is relatively well established across most major package managers, and (b) as long as you define the meanings, it doesn't really matter what others do.

An analogous argument would be "why doesn't your programming language have a string concat operator"; "because other languages use different characters for string concat operators."

500-CNAB-dependencies.md
```diff
@@ -0,0 +1,100 @@
...
+* This doesn't say anything about how to use the dependency!
+    * This requires that the [outputs PR](https://github.com/deislabs/cnab-spec/pull/133) is merged. Once we have that, then given
+        a named bundle, we know exactly what the set of outputs of the bundle
+        will be and how to consume them.
+* Why didn't you fill out the "provides/requires" idea that you proposed a few weeks ago?
+    * This is a more straightforward idea for us to argue about first. 😅 We can
+        bikeshed over if it's a seperate spec, and on so many other wonderful topics
+        and on a very easy problem. Then if it's merged, then we can try tackling
+        the much more difficult problem of [Depend on a bundle that provides a capability](#depend-on-a-bundle-that-provides-a-capability).
```
technosophos: This is a very good approach. That whole virtual dependency thing was a nightmare in Debian, then a nightmare again in Gentoo. Not trying to figure it out in the rough drafts is probably good.

500-CNAB-dependencies.md
technosophos: Just to be clear, I'm not suggesting that we MUST have those operators. I'm merely saying that if that is the only reason, I don't find it to be a good one.

The more I think about it, the more I think this is actually not something you even have to address in the spec. If the .x notation works for what you need, just don't bother distracting people with this discussion.

carolynvs: Per the comment under the heading, these Q&A's aren't going into the spec. They are just a place for me to proactively ask and answer expected questions about what I did in this proposal.

500-CNAB-dependencies.md
```diff
@@ -0,0 +1,100 @@
...
+
+This specification introduces a `dependencies` array in the bundle.json
+that defines metadata necessary to specify a dependency.
+
+### requires
+
+The `requires` object defines the criteria for the dependent bundle:
+
+* `bundle`: A reference to a bundle in the format REGISTRY/NAME.
+* `version`: A set of criteria applied to the bundle version when selecting an
```
silvin-lubecki: I have a question, which may be dummy, about the versions and the reference to a bundle in the format `REGISTRY/NAME`:
Which version are we talking about? The one defined in the bundle.json, or the tag to reference the right bundle version in a registry `REGISTRY/NAME:TAG` ?

My point is:
- if we use the `version` field in the bundle, then the resolver will need to pull all the bundles, parse them, then resolve the best version. I guess this can become heavy with time and multiple versions.
- if we use the `tag`, the resolver has just to fetch all the tags to select the best version, but then it means we must tight the tag with the version field, as there is no way to guarantee that the tag is relevant with its version.

Does my question make sense? Am I missing something?

carolynvs: My  intention was to use the latter, where the bundle author is expected to tag their bundles with the bundle version so that people using this dependency spec can depend upon this convention.

Do you anticipate a problem with that or is that a reasonable convention that people can follow?

michelleN: The registry conversation and how duffle pushes bundles to registries right now makes it likely that bundles will be tagged with a digest and not semver. What is your thinking around referring to a bundle name with its associated tag all in the `BUNDLE` field? Perhaps that misses the point since you'd have to create a new bundle for every dependency bump.

silvin-lubecki: > it likely that bundles will be tagged with a digest and not semver.

With `docker app` we push the bundles with a tag, and use `semver` by default (even if the user can choose any tag if she wants), and I guess most of the CNAB tools will do the same?

> Do you anticipate a problem with that or is that a reasonable convention that people can follow?

No that's a perfectly reasonable convention 👍 I don't anticipate a problem, just wanted to make sure we were talking about using `tags` for the dependency management. That said, maybe it should be added in the specifications? Or maybe is it purely an implementation detail?

PR #167, opened 2019-04-26
add Urvashi as core maintainer

michelleN:
I'm pleased to announce that @youreddy is now a core maintainer on the CNAB spec. 🎉


michelleN:
I just need one more lgtm here to merge cc/ @technosophos

PR #174, opened 2019-05-09
Adds missing writeOnly parameter definition

ryanmoran:
We missed the `writeOnly` field when we first landed JSONSchema support in parameters. This PR remedies that.

cc @youreddy

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -442,7 +442,7 @@ The structure of a parameters section looks like the section below.
       "then": <json-schema>,
       "title": <string>,
       "type": <string> | [ <string> ],
-      "uniqueItems": boolean,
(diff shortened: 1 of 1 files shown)


101-bundle-json.md
```diff
@@ -441,8 +442,8 @@ The structure of a parameters section looks like the section below.
       "then": <json-schema>,
       "title": <string>,
       "type": <string> | [ <string> ],
-      "uniqueItems": boolean
-
+      "uniqueItems": boolean,
```
michelleN: should this be `"uniqueItems": <boolean>`

ryanmoran: Good catch! I've fixed that now.

michelleN: awesome thanks!

PR #186, opened 2019-06-06
change apply-to to applyTo

### LINKED ISSUE #149: Rename `apply-to` to `applyTo`
I just noticed that `apply-to` is the only property that is not camel cased. It should be renamed to `applyTo` to match everything else.

technosophos:
To match casing elsewhere, rename apply-to to applyTo.

Closes #149


jeremyrickard:
Thanks @technosophos !

PR #187, opened 2019-06-06
remove mention of drivers, which are specific to duffle

### LINKED ISSUE #180: Delete all mentions of drivers
According to the discussion in the [CNAB community meeting of 27 Feb 2019](https://hackmd.io/s/SyGcBcwQ4#February-27-Agenda):
> [...] drivers are part of the duffle implementation and are not part of the spec.

So all references to drivers in the spec should be deleted.

This should include the following paragraph from [invocation images](https://github.com/deislabs/cnab-spec/blob/0abbb9e0d52b0ae631d12cf8278711c69fb24c64/101-bundle-json.md#invocation-images):

> Duffle, the reference implementation of a CNAB installer, introduces a layer of user-customizable drivers which are type-aware. Images MAY be delegated to drivers for installation.

technosophos:
This changes the language in several places. Instead of suggesting the _driver_ determines which invocation image is run, it now reads as the _CNAB runtime_ must select the driver.

Further, I added a clarification that if no sufficient image is found, the runtime must return an error.

Closes #180

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -251,7 +251,7 @@ The following fields are informational pieces of metadata designed to convey add
 
 ## Invocation Images
 
-The `invocationImages` section describes the images that contain the bootstrapping for the image. The appropriate invocation image is selected by the CNAB runtime, typically by considering the runtime requirements of the bundle. For example, both a Windows and a Linux version of the invocation image may be included in the list. It is up to the CNAB runtime to determine which one to use. If no sufficient image is found, the CNAB runtime MUST emit an error and stop processing.
+The `invocationImages` section describes the images that are responsible for bootstrapping the installation. The appropriate invocation image is selected by the CNAB runtime, typically by considering the runtime requirements of the bundle. For example, both a Windows and a Linux version of the invocation image may be included in the list. It is up to the CNAB runtime to determine which one to use. If no sufficient image is found, the CNAB runtime MUST emit an error and stop processing.
 
 A CNAB bundle MUST have at least one invocation image.
 
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -251,7 +251,7 @@ The following fields are informational pieces of metadata designed to convey add
 
 ## Invocation Images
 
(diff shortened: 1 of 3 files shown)


101-bundle-json.md
```diff
@@ -251,7 +251,7 @@ The following fields are informational pieces of metadata designed to convey add
 
 ## Invocation Images
 
-The `invocationImages` section describes the images that contains the bootstrapping for the image. The appropriate invocation image is selected using the current driver.
+The `invocationImages` section describes the images that contains the bootstrapping for the image. The appropriate invocation image is selected by the CNAB runtime, typically by considering the runtime requirements of the bundle. For example, both a Windows and a Linux version of the invocation image may be included in the list. It is up to the CNAB runtime to determine which one to use. If no sufficient image is found, the CNAB runtime MUST emit an error and stop processing.
```
jlegrone:
```suggestion
The `invocationImages` section describes the images that contain the bootstrapping for the image. The appropriate invocation image is selected by the CNAB runtime, typically by considering the runtime requirements of the bundle. For example, both a Windows and a Linux version of the invocation image may be included in the list. It is up to the CNAB runtime to determine which one to use. If no sufficient image is found, the CNAB runtime MUST emit an error and stop processing.
```

technosophos:
Good catch, @jlegrone . Applied the patch. Thanks,

101-bundle-json.md
```diff
@@ -251,7 +251,7 @@ The following fields are informational pieces of metadata designed to convey add
 
 ## Invocation Images
 
-The `invocationImages` section describes the images that contains the bootstrapping for the image. The appropriate invocation image is selected using the current driver.
+The `invocationImages` section describes the images that contain the bootstrapping for the image. The appropriate invocation image is selected by the CNAB runtime, typically by considering the runtime requirements of the bundle. For example, both a Windows and a Linux version of the invocation image may be included in the list. It is up to the CNAB runtime to determine which one to use. If no sufficient image is found, the CNAB runtime MUST emit an error and stop processing.
```
jlegrone: Actually I just read this again and the first sentence is kind of hard to parse IMO. While we're at it, maybe something like this would be better?

```suggestion
The `invocationImages` section describes the images that are responsible for bootstrapping the installation. The appropriate invocation image is selected by the CNAB runtime, typically by considering the runtime requirements of the bundle. For example, both a Windows and a Linux version of the invocation image may be included in the list. It is up to the CNAB runtime to determine which one to use. If no sufficient image is found, the CNAB runtime MUST emit an error and stop processing.
```

radu-matei: Yeah, I think the suggestion makes it a bit easier to understand.

glyn:
Great stuff - thanks!

PR #190, opened 2019-06-06
Added some wording to the specification regarding output paths

### LINKED ISSUE #184: Proposal: Require output paths to be subpaths of /cnab/app/outputs
Right now the spec does not specify a location for [output paths](https://github.com/deislabs/cnab-spec/blob/0abbb9e0d52b0ae631d12cf8278711c69fb24c64/101-bundle-json.md#outputs). This can make driver implementation tricky in some cases, and can also lead to conflicts with params and credentials.

I propose modifying the spec to say the following:

1. An output path MUST be a subpath of `/cnab/app/outputs`
2. Param or credential paths MUST NOT be subpaths of `/cnab/app/outputs`

jeremyrickard:
* Adds specification to the output section that outputs must be under /cnab/app/outputs
* Adds specification to params/creds that they must NOT be under /cnab/app/outputs

cc: @jlegrone

Closes #184

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -381,7 +381,7 @@ Parameter specifications consist of name/value pairs. The name is fixed, but the
       - `description`: Descriptive text for the field. Can be used to decorate a user interface. MUST be a string. (OPTIONAL)
       - `destination`: Indicates where (in the invocation image) the parameter is to be written (REQUIRED)
         - `env`: The name of an environment variable
-        - `path`: The fully qualified path to a file that will be created. Specified path MUST NOT be a subpath of /cnab/app/outputs.
+        - `path`: The fully qualified path to a file that will be created. Specified path MUST NOT be a subpath of `/cnab/app/outputs`.
       - `else`: Parameter validation requiring that the user-provided value match the specified schema. Only matches if the user-provided value does NOT match the schema provided in the `if` property. MUST be a JSON schema. (OPTIONAL)
       - `enum`: Parameter validation requiring that the user-provided value is one of the specified items in the specified array. MUST be a non-empty array of unique elements that can be of any type. (OPTIONAL)
       - `examples`: Sample JSON values associated with a particular schema. MUST be an array. (OPTIONAL)
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -381,7 +381,7 @@ Parameter specifications consist of name/value pairs. The name is fixed, but the
       - `description`: Descriptive text for the field. Can be used to decorate a user interface. MUST be a string. (OPTIONAL)
       - `destination`: Indicates where (in the invocation image) the parameter is to be written (REQUIRED)
         - `env`: The name of an environment variable
-        - `path`: The fully qualified path to a file that will be created
+        - `path`: The fully qualified path to a file that will be created. Specified path MUST NOT be a subpath of /cnab/app/outputs.
       - `else`: Parameter validation requiring that the user-provided value match the specified schema. Only matches if the user-provided value does NOT match the schema provided in the `if` property. MUST be a JSON schema. (OPTIONAL)
       - `enum`: Parameter validation requiring that the user-provided value is one of the specified items in the specified array. MUST be a non-empty array of unique elements that can be of any type. (OPTIONAL)
       - `examples`: Sample JSON values associated with a particular schema. MUST be an array. (OPTIONAL)
@@ -646,7 +646,7 @@ What about parameters such as database passwords used by the application? Proper
 
 - The `credentials` container is a map of human-friendly credential names to a description of where the invocation image expects to find them.
   - The name key MUST be human-readable
-    - `path` describes the _absolute path within the invocation image_ where the invocation image expects to find the credential
+    - `path` describes the _absolute path within the invocation image_ where the invocation image expects to find the credential. Specified path MUST NOT be a subpath of /cnab/app/outputs.
     - `env` contains _the name of an environment variable_ that the invocation image expects to have available when executing the CNAB `run` tool (covered in the next section).
     - `description` contains a user-friendly description of the credential.
     - `required` indicates whether this credential MUST be supplied. By default, it is `false`, which means the credential is optional. When `true`, a runtime MUST fail if the credential is not provided.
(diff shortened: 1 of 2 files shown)


101-bundle-json.md
```diff
@@ -377,7 +377,7 @@ Parameter specifications consist of name/value pairs. The name is fixed, but the
     - `description`: Descriptive text for the field. Can be used to decorate a user interface. MUST be a string. (OPTIONAL)
     - `destination`: Indicates where (in the invocation image) the parameter is to be written (REQUIRED)
       - `env`: The name of an environment variable
-      - `path`: The fully qualified path to a file that will be created
+      - `path`: The fully qualified path to a file that will be created. Specified path MUST NOT be a subpath of /cnab/app/outputs.
```
technosophos: Good catch!

radu-matei: nit: can you please format the path?

101-bundle-json.md
```diff
@@ -633,7 +633,7 @@ What about parameters such as database passwords used by the application? Proper
 
 - The `credentials` container is a map of human-friendly credential names to a description of where the invocation image expects to find them.
   - The name key MUST be human-readable
-    - `path` describes the _absolute path within the invocation image_ where the invocation image expects to find the credential
+    - `path` describes the _absolute path within the invocation image_ where the invocation image expects to find the credential. Specified path MUST NOT be a subpath of /cnab/app/outputs.
```
radu-matei: nit: can you please format the path?

101-bundle-json.md
```diff
@@ -802,7 +802,7 @@ Output specifications are flat (not tree-like), consisting of name/value pairs.
     - `multipleOf`: Output validation requiring that the resulting number be wholly divisible by the number specified. MUST be a number strictly greater than zero. (OPTIONAL)
     - `not`: Output validation requiring that the resulting value NOT match the specified schema. MUST be a JSON schema. (OPTIONAL)
     - `oneOf`: Output validation requiring that the resulting value match ONE of the specified schemas. MUST be a non-empty array of JSON schemas. (OPTIONAL)
-    - `path`: The fully qualified path to a file that will be created (REQUIRED)
+    - `path`: The fully qualified path to a file that will be created (REQUIRED). The path specified MUST be a subpath of /cnab/app/outputs.
```
glyn: I think the path needs to be a strict subpath of `/cnab/app/outputs`. If the path is equal to `/cnab/app/outputs`, something bad may happen.

glyn: Do we allow distinct outputs to use the same `path` value? I hope not. Probably worth ruling this out with some "MUST NOT" wording.

schema/bundle.schema.json
```diff
@@ -286,7 +288,8 @@
                 },
                 "path":{
                     "description": "The path inside of the invocation image where output will be written",
-                    "type": "string"
+                    "type": "string",
+                    "pattern": "^\/cnab\/app\/outputs\/.+$"
```
glyn: Good. If I'm reading this correctly, it specifies a strict subpath of `/cnab/app/outputs`.

jeremyrickard: Yeah, I added the wording in the prose to clarify this :)

101-bundle-json.md
jeremyrickard: I added that this path must be unique.

jeremyrickard: Yeah, that was the intent. Added _strict_ in front of subpath.

technosophos:
I think this just needs a re-review from @glyn and then it can be merged, right?

101-bundle-json.md
```diff
@@ -802,7 +802,7 @@ Output specifications are flat (not tree-like), consisting of name/value pairs.
     - `multipleOf`: Output validation requiring that the resulting number be wholly divisible by the number specified. MUST be a number strictly greater than zero. (OPTIONAL)
     - `not`: Output validation requiring that the resulting value NOT match the specified schema. MUST be a JSON schema. (OPTIONAL)
     - `oneOf`: Output validation requiring that the resulting value match ONE of the specified schemas. MUST be a non-empty array of JSON schemas. (OPTIONAL)
-    - `path`: The fully qualified path to a file that will be created (REQUIRED)
+    - `path`: The fully qualified path to a file that will be created (REQUIRED). The path specified MUST be a _strict_ subpath of `/cnab/app/outputs` and MUST be unique for this output.
```
glyn: That isn't quite right. Of course, the path will be unique for this output because it's the only path for this output. I think the wording needs to be something like: "and MUST be distinct from the paths for all other outputs in this bundle.".

jeremyrickard:
@glyn I'll work on updating the language again on Monday.

 I'd like to ask that in the future that we open up follow-up issues for the sort of thing you're addressing right now. Your comment about uniqueness wasn't really related to the _specific_ change this PR introduced and now getting _that_ specific change is prolonged because we are nit-ing the langauge of a not directly related change.

glyn:
> @glyn I'll work on updating the language again on Monday.

Thanks.

> 
> I'd like to ask that in the future that we open up follow-up issues for the sort of thing you're addressing right now. Your comment about uniqueness wasn't really related to the _specific_ change this PR introduced and now getting _that_ specific change is prolonged because we are nit-ing the langauge of a not directly related change.

Hmmm, I'm not sure I follow. I thought I was asking for a tightening up of some text introduced by this PR, so I don't really understand why this wasn't really related to the specific change this PR introduced. Perhaps it would have been more constructive to add a suggestion which you could have instantly merged. I'll go back and see if I can do that.

101-bundle-json.md
glyn:
```suggestion
    - `path`: The fully qualified path to a file that will be created (REQUIRED). The path specified MUST be a _strict_ subpath of `/cnab/app/outputs` and MUST be distinct from the paths for all other outputs in this bundle.
```

glyn: Added a suggestion instead.

jeremyrickard:
@glyn please re-review 

cc: @technosophos @simonferquel

ndeloof:
I went into some troubles with the cnab spec when trying to validate a bundle.json in Go. Regular expression `^(?!\/cnab\/app\/outputs)\/.+$` uses a lookahead assertion, which isn't supported by Go regex standard library (see https://groups.google.com/forum/#!topic/golang-nuts/7qgSDWPIh_E). This prevents using  https://github.com/xeipuuv/gojsonschema for validation.

I know https://github.com/deislabs/cnab-go do offer `bundle.Validate()` but this one won't validate against the specification schema.

PR #192, opened 2019-06-11
Factors JSONSchema definitions out of parameters

ryanmoran:
- these schema are now defined by name in a top-level field called
  "definitions"
- parameters will define the schema for their field by referencing the
  name of a definition in their own "definition" field

This did not touch the `outputs` section, but it would make sense to also factor the validation for those fields out to this same top-level `definitions` field. That would separate the concerns of validation from those of naming and exposing those user inputs during invocation.

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -20,10 +20,11 @@ A `bundle.json` is broken down into the following categories of information:
   - `description`: Short description of the bundle
 - Information on the invocation images, as an array
 - A map of images included with this bundle, as a `component name` to `image definition` map
-- A specification of which parameters MAY be overridden, and how those are to be validated
+- A specification of which parameters MAY be overridden, and a reference to a validation schema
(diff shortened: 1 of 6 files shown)


ryanmoran:
cc: @jeremyrickard

technosophos:
I'm on board with this, but will leave it to @jeremyrickard to do the second formal LGTM. What's the issue number that this one closes? I seem to recall reading it, but now I can't find it.

ryanmoran:
@technosophos the issue that contained the original proposal is https://github.com/deislabs/cnab-spec/issues/178.

technosophos:
Ah! Thanks! I forgot about that original title!

jeremyrickard:
@ryanmoran LGTM, will approve once you resolve the conflict (sorry I merged first :()

ryanmoran:
@jeremyrickard just fixed those conflicts.

PR #203, opened 2019-06-21
Revisions to 100-CNAB and 101-bundle-json

astrieanna:
In working to understand the CNAB spec, I started reading at the beginning and reading the documents very closely. When documents grow over time via PR and are in a state of flux as these are, they can become a little tough to read.

This PR presents some edits that make some sentences read more smoothly to me, more sentences into paragraphs on the same topic, and restructure the definitions of some terms (thick/thin bundles). My intention is not to change the meaning of the spec with the edits, but I may have misunderstood the meaning.

There are two more adventurous edits in 101-bundle-json. The "unicode letters" and "unicode graph characters" seemed like very vague definitions, so I gave them my best guess at a precise definition. These definitions mapped well the the existing phrases and seemed like not unreasonable choices for the spec to have made, but if this isn't what you meant I'm happy to change it.
 
* "unicode letters" became "any unicode code point with a General Category of Lu, Ll, Lt, Lm, or Lo (a General Major Category of Letter)"
* "unicode graph characters" became "Unicode code points with a basic type of `Graphic`"

cc @ryanmoran

### DIFF
--- a/100-CNAB.md
+++ b/100-CNAB.md
@@ -7,78 +7,87 @@ weight: 100
 *[Working Draft](901-process.md), Nov. 2018*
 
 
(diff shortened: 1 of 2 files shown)


astrieanna:
I'm going to split this up to make it easier to discuss and review.

PR #212, opened 2019-07-01
Add immutable flag to parameters.

### LINKED ISSUE #181: Proposal:Add an immutable property to bundle parameter object
Some parameters to bundles may be immutable, examples include:

- properties of resources that can only be set when those resources are created (e.g. in Azure a VM password can only be set when a VM is created, attempts to update it will be met with failure)

- where a bundle targets new or existing infrastructure resources its important to know on subsequent actions if those infrastructure resources were created as part of the application install (e.g. if I create a resource as part of an bundle install then I need to know on uninstall in order to handle the cleanup correctly) 

Proposal is to add a property **immutable** to the parameter object with a default setting of  false, immutable properties would only be able to be set on the install action,

jeremyrickard:
This PR adds an immutable flag to the parameter definition in `101-bundle-json.md` and adds wording to `103-bundle-runtime.md` to indicate that if an immutable parameter value is given at any time other than runtime, it should result in an error. 

Closes #181

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -233,7 +233,8 @@ And here is how a "thick" bundle looks. Notice how the `invocationImage` and `im
         "description": "The port that the backend will listen on",
         "destination": {
           "path": "/path/to/backend_port"
-        }
+        },
+        "immutable" : true
       }
     }
   },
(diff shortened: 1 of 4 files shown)


technosophos:
Does the JSON Schema need an update?

jeremyrickard:
@technosophos indeed it does! Sorry I thought I opened this as a Draft PR. working on the schema still :(

jeremyrickard:
@technosophos Updated the PR to add the schema addition and updated the 101.02 to add an explicit `"immutable": true` entry.

PR #216, opened 2019-07-09
Fix dependency example json

carolynvs:
* When I went to [add support for dependencies in cnab-go](https://github.com/deislabs/cnab-go/pull/55), I realized that the example json wasn't using array/object properly. This is what I get for writing json embedded in markdown... 🤦‍♀️
* Use bool instead of bool string for prereleases

**Update:** Since this isn't part of the core spec, it can't live at the root because that messes up the json schema valiation. So it will live under the custom extensions, custom.

* Add a json schema document
* Add a validation test for the schema doc

### DIFF
--- a/500-CNAB-dependencies.md
+++ b/500-CNAB-dependencies.md
@@ -38,18 +38,17 @@ stored in the custom extensions section of the bundle.
 {
   "custom": {
     "dependencies": {
-      "requires": [
-        {
+      "requires": {
+        "storage": {
           "bundle": "somecloud/blob-storage"
         },
-        {
+        "mysql": {
           "bundle": "somecloud/mysql",
           "version": {
             "prereleases": true,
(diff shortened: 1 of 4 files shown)


carolynvs:
I've updated this PR to move "dependencies" under "custom". Since this isn't part of the core spec, it can't live at the root because that messes up the json schema valiation. So it will live under the custom extensions, custom.

* Add a json schema document
* Add a validation test for the schema doc

ryanmoran:
Do we want to "reserve" the `dependencies` key in the `custom` object?

carolynvs:
That was @jeremyrickard and my intention.

technosophos:
Do we need to explicitly state that? Or can we merge this PR as-is?

carolynvs:
I'll update it to state that dependencies is reserved.

PR #220, opened 2019-07-10
Adding "requiredExtensions" to core spec

jeremyrickard:
This PR introduces a `requiredExtensions` object and modifies
the wording of 101-bundle-json.md to refine the existing custom
extensions section. Relevant example and JSON schema changes included.

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -856,7 +856,7 @@ The fields are defined as follows:
 
 Some extensions MAY be required by a bundle and operations involving that bundle MUST fail if the runtime does not support that bundle. In this case, a bundle author MUST use the `requiredExtensions` object to define those extensions that are necessary for installing the bundle. The `requiredExtensions` object should contain a collection of key/value pairs. The key MUST be the `EXTENSION NAME` and the value MUST be a well known JSON Schema reference that can be used to validate the JSON data contained in the `custom` object.
 
-A runtime MUST validate that it supports any defined custom actions before performing any operation. However, bundles SHOULD be installable by runtimes that do not understand the extensions if they are NOT identified in the `requiredExtensions` field.  
+A runtime SHOULD check that it supports any required custom actions before performing any operation. However, bundles SHOULD be installable by runtimes that do not support extensions that are NOT identified in the `requiredExtensions` field.  
 
 See [101.03-bundle.json](examples/101.03-bundle.json) for a complete example.
 
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -854,7 +854,7 @@ The fields are defined as follows:
 
 ### Required Extensions
 
-Some extensions MAY be required by a bundle and operations involving that bundle MUST fail if the runtime does not support that bundle. In this case, a bundle author MUST use the `requiredExtensions` array to define those extensions that are necessary for installing the bundle. The `requiredExtensions` array should contain the `EXTENSION NAME` defined in the `custom` object for any extensions that are required for bundle operations.
+Some extensions MAY be required by a bundle and operations involving that bundle MUST fail if the runtime does not support that extension. In this case, a bundle author MUST use the `requiredExtensions` array to define those extensions that are necessary for installing the bundle. The `requiredExtensions` array should contain the `EXTENSION NAME` defined in the `custom` object for any extensions that are required for bundle operations.
 
 A runtime SHOULD check that it supports any required custom actions before performing any operation. However, bundles SHOULD be installable by runtimes that do not support extensions that are NOT identified in the `requiredExtensions` field.  
 
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -854,7 +854,7 @@ The fields are defined as follows:
 
 ### Required Extensions
 
-Some extensions MAY be required by a bundle and operations involving that bundle MUST fail if the runtime does not support that extension. In this case, a bundle author MUST use the `requiredExtensions` array to define those extensions that are necessary for installing the bundle. The `requiredExtensions` array should contain the `EXTENSION NAME` defined in the `custom` object for any extensions that are required for bundle operations.
+Some extensions defined in the `custom` object of a bundle MAY be required in order for a runtime to perform any action on the bundle. A bundle author MUST use the `requiredExtensions` array to define those extensions that are required. The `requiredExtensions` array SHOULD contain the `EXTENSION NAME` defined in the `custom` object for each extension that is required.
 
 A runtime SHOULD check that it supports any required custom actions before performing any operation. However, bundles SHOULD be installable by runtimes that do not support extensions that are NOT identified in the `requiredExtensions` field.  
 
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -852,13 +852,19 @@ The fields are defined as follows:
   - `EXTENSION NAME`: a unique name for an extension. Names SHOULD follow the dotted name format described earlier in this section.
   - The value of the extension must be valid JSON, but is otherwise undefined.
 
-The usage of extensions is undefined. However, bundles SHOULD be installable by runtimes that do not understand the extensions.
+### Required Extensions
+
+Some extensions MAY be required by a bundle and operations involving that bundle MUST fail if the runtime does not support that bundle. In this case, a bundle author MUST use the `requiredExtensions` object to define those extensions that are necessary for installing the bundle. The `requiredExtensions` object should contain a collection of key/value pairs. The key MUST be the `EXTENSION NAME` and the value MUST be a well known JSON Schema reference that can be used to validate the JSON data contained in the `custom` object.
+
+A runtime MUST validate that it supports any defined custom actions before performing any operation. However, bundles SHOULD be installable by runtimes that do not understand the extensions if they are NOT identified in the `requiredExtensions` field.  
+
+See [101.03-bundle.json](examples/101.03-bundle.json) for a complete example.
 
 ## Outputs
 
 The `outputs` section of the `bundle.json` defines which outputs an application will produce during the course of executing a bundle. Outputs are expected to be written to one or more files on the file system of the invocation image. The location of this file MUST be provided in the output definition.
 
-Output specifications are flat (not tree-like), consisting of name/value pairs. The output definition includes a destination the output will be written to, along with a definition to help validate their contents.
+Output specifications are flat (not tree-like), consisting of name/value pairs. The output definition includes a destination the output will be written to, along with a definition to help validate their contents. 
 
 ```json
 {
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -856,7 +856,7 @@ The fields are defined as follows:
 
 Some extensions defined in the `custom` object of a bundle MAY be required in order for a runtime to perform any action on the bundle. A bundle author MUST use the `requiredExtensions` array to define those extensions that are required. The `requiredExtensions` array SHOULD contain the `EXTENSION NAME` defined in the `custom` object for each extension that is required.
 
-A runtime SHOULD check that it supports any required custom actions before performing any operation. However, bundles SHOULD be installable by runtimes that do not support extensions that are NOT identified in the `requiredExtensions` field.  
+A runtime MUST check that it supports any required extensions before performing any action on the bundle. If the runtime does not support the required extension(s), it MAY proceed with the action or fail, however it MUST notify the user that it does not support the required extension(s). Runtimes that do not support extensions that are NOT identified in the `requiredExtensions` field of a bundle SHOULD perform actions on the bundle. 
 
 ```json
 {
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -854,7 +854,7 @@ The fields are defined as follows:
 
 ### Required Extensions
 
-Some extensions MAY be required by a bundle and operations involving that bundle MUST fail if the runtime does not support that bundle. In this case, a bundle author MUST use the `requiredExtensions` array to define those extensions that are necessary for installing the bundle. The `requiredExtensions` object should contain the `EXTENSION NAME` defined in the `custom` object for any extensions that are required for bundle operations.
+Some extensions MAY be required by a bundle and operations involving that bundle MUST fail if the runtime does not support that bundle. In this case, a bundle author MUST use the `requiredExtensions` array to define those extensions that are necessary for installing the bundle. The `requiredExtensions` array should contain the `EXTENSION NAME` defined in the `custom` object for any extensions that are required for bundle operations.
 
 A runtime SHOULD check that it supports any required custom actions before performing any operation. However, bundles SHOULD be installable by runtimes that do not support extensions that are NOT identified in the `requiredExtensions` field.  
 
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -858,7 +858,119 @@ Some extensions MAY be required by a bundle and operations involving that bundle
 
 A runtime SHOULD check that it supports any required custom actions before performing any operation. However, bundles SHOULD be installable by runtimes that do not support extensions that are NOT identified in the `requiredExtensions` field.  
 
-See [101.03-bundle.json](examples/101.03-bundle.json) for a complete example.
+```json
+{
+  "credentials": {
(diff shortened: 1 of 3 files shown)


101-bundle-json.md
```diff
@@ -852,13 +852,19 @@ The fields are defined as follows:
   - `EXTENSION NAME`: a unique name for an extension. Names SHOULD follow the dotted name format described earlier in this section.
   - The value of the extension must be valid JSON, but is otherwise undefined.
 
-The usage of extensions is undefined. However, bundles SHOULD be installable by runtimes that do not understand the extensions.
+### Required Extensions
+
+Some extensions MAY be required by a bundle and operations involving that bundle MUST fail if the runtime does not support that bundle. In this case, a bundle author MUST use the `requiredExtensions` object to define those extensions that are necessary for installing the bundle. The `requiredExtensions` object should contain a collection of key/value pairs. The key MUST be the `EXTENSION NAME` and the value MUST be a well known JSON Schema reference that can be used to validate the JSON data contained in the `custom` object.
+
+A runtime MUST validate that it supports any defined custom actions before performing any operation. However, bundles SHOULD be installable by runtimes that do not understand the extensions if they are NOT identified in the `requiredExtensions` field.
```
glyn: 1. Weaken "MUST" to "SHOULD" for consistency with the above comment on allowing a user to "force" an operation.

2. Perhaps we should avoid the word "validate" here as it might imply just schema validation.

3. We should talk about "required" actions rather than "defined" actions.

4. Let's use "support" rather than "understand" for consistency.
```suggestion
A runtime SHOULD check that it supports any required custom actions before performing any operation. However, bundles SHOULD be installable by runtimes that do not support extensions that are NOT identified in the `requiredExtensions` field.  
```

101-bundle-json.md
```diff
@@ -852,13 +852,19 @@ The fields are defined as follows:
...
   - The value of the extension must be valid JSON, but is otherwise undefined.
 
-The usage of extensions is undefined. However, bundles SHOULD be installable by runtimes that do not understand the extensions.
+### Required Extensions
+
+Some extensions MAY be required by a bundle and operations involving that bundle MUST fail if the runtime does not support that bundle. In this case, a bundle author MUST use the `requiredExtensions` object to define those extensions that are necessary for installing the bundle. The `requiredExtensions` object should contain a collection of key/value pairs. The key MUST be the `EXTENSION NAME` and the value MUST be a well known JSON Schema reference that can be used to validate the JSON data contained in the `custom` object.
+
+A runtime MUST validate that it supports any defined custom actions before performing any operation. However, bundles SHOULD be installable by runtimes that do not understand the extensions if they are NOT identified in the `requiredExtensions` field.  
+
+See [101.03-bundle.json](examples/101.03-bundle.json) for a complete example.
```
glyn: Elsewhere, we inline the example and follow it with "Source: [link]".

examples/101.03-bundle.json
```diff
@@ -0,0 +1,110 @@
...
+        }
+      }
+    },
+    "requiredExtensions" : {
+        "dependencies" : "cnab.io/specs/depenencies.schema.json"
+    },
+    "schemaVersion": "v1.0.0-WD",
+    "version": "0.1.2"
+  }
+
```
glyn: Nit: newline problem.

examples/101.03-bundle.json
```diff
@@ -0,0 +1,110 @@
...
+          "definition": "http_port",
+          "description": "The port that the back-end will listen on",
+          "destination": {
+            "env": "BACKEND_PORT"
+          }
+        }
+      }
+    },
+    "requiredExtensions" : {
+        "dependencies" : "cnab.io/specs/depenencies.schema.json"
```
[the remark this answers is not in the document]
jeremyrickard: JSON Schema doesn't require that any validation tools need to make any network calls and this is just a reference. We wouldn't actually be able to pull these down in any sort of disconnected mode of operations either, so I'm not sure that HTTPS for security really gives anything? 

I think for most extensions there will be some required functionality that needs to be handled ahead of time, so I think they are kind of "registered" with the tooling provider at least and this should just be a "well known" extension schema?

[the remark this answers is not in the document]
glyn: Seems reasonable, but I'm still not clear what "well known" really means. For example, suppose a company is implementing their own custom extension (for use only by the company's runtime and bundles -- no-one else knows are cares about the extension). Does the extension need to be "published" in order to be "well known" or can it be kept private?

101-bundle-json.md
```diff
@@ -852,13 +852,19 @@ The fields are defined as follows:
   - `EXTENSION NAME`: a unique name for an extension. Names SHOULD follow the dotted name format described earlier in this section.
   - The value of the extension must be valid JSON, but is otherwise undefined.
 
-The usage of extensions is undefined. However, bundles SHOULD be installable by runtimes that do not understand the extensions.
+### Required Extensions
+
+Some extensions MAY be required by a bundle and operations involving that bundle MUST fail if the runtime does not support that bundle. In this case, a bundle author MUST use the `requiredExtensions` object to define those extensions that are necessary for installing the bundle. The `requiredExtensions` object should contain a collection of key/value pairs. The key MUST be the `EXTENSION NAME` and the value MUST be a well known JSON Schema reference that can be used to validate the JSON data contained in the `custom` object.
```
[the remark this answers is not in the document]
jeremyrickard: I'm good with all of this, except:

> I'm not sure what "well known" means, but I think we should require a URL for the schema reference as the runtime needs to be able to locate (and "download") the schema not just know it's identifier.

I don't think we can assume that a tool can ever "download" the schema at runtime. In the case of running the tool in a disconnected/offline environment, that wouldn't be possible. Most extensions are very likely to require some runtime capabilities (dependencies for example) that need to be included in the tool (it's possible some tools could dynamically load extensions but let's assume they don't for simplicity) and known ahead of time. That's what was implied by the "well known" aspect of it. I think we can absolutely refer to the _URL_ of a schema for an extension but I think we should not assume that it's going to be downloaded. Thoughts?

examples/101.03-bundle.json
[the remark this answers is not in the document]
jeremyrickard: I think we can remove the well known part? I'm a private company building my own extension that I don't plan on publishing (sad :() but I'd still reference it here with some sort of URI like reference yes? 

So maybe just remove the "well known" part? That was really trying to address that a bundle author could reference a published (i.e. well known) extension (dependencies for example) and a tool could validate that it supports that extension because it knows the schema reference)

101-bundle-json.md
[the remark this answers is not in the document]
glyn: Yes, that seems reasonable. So "well known" kind of means "globally unique" in that it somehow won't clash with the schema reference of another extension. Right?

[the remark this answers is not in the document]
jeremyrickard: I think that's a true statement. And that "cnab.io/spec/extensions/dependencies" refers to dependencies and that a bundle author can say "i need that extension" and tools that support dependencies will work with that reference. Maybe we don't really need the reference, maybe we need agreed upon extension names and it's just assumed that tooling will know that extension name and support it or not. I think that's really what I was trying to get at with that.

jeremyrickard:
Thanks for hashing this out with me @glyn!

examples/101.03-bundle.json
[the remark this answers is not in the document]
glyn: It seems like the schema reference needs to be unique across extensions but doesn't need to refer to a JSON schema at all. If true, I find that confusing. I'm wondering what the difference in function is between a schema reference and the reverse DNS extension name.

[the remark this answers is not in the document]
jeremyrickard: Following up on our slack conversation:

Proposal: "requiredExtensions" just becomes an array of "EXTENSION NAME" in custom.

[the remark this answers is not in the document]
glyn: Agreed, thanks.

101-bundle-json.md
```diff
@@ -852,13 +852,131 @@ The fields are defined as follows:
   - `EXTENSION NAME`: a unique name for an extension. Names SHOULD follow the dotted name format described earlier in this section.
   - The value of the extension must be valid JSON, but is otherwise undefined.
 
-The usage of extensions is undefined. However, bundles SHOULD be installable by runtimes that do not understand the extensions.
+### Required Extensions
+
+Some extensions MAY be required by a bundle and operations involving that bundle MUST fail if the runtime does not support that bundle. In this case, a bundle author MUST use the `requiredExtensions` array to define those extensions that are necessary for installing the bundle. The `requiredExtensions` object should contain the `EXTENSION NAME` defined in the `custom` object for any extensions that are required for bundle operations.
```
glyn:
```suggestion
Some extensions MAY be required by a bundle and operations involving that bundle MUST fail if the runtime does not support that bundle. In this case, a bundle author MUST use the `requiredExtensions` array to define those extensions that are necessary for installing the bundle. The `requiredExtensions` array should contain the `EXTENSION NAME` defined in the `custom` object for any extensions that are required for bundle operations.
```

examples/101.03-bundle.json
```diff
@@ -0,0 +1,109 @@
...
+        "definition": "http_port",
+        "description": "The port that the back-end will listen on",
+        "destination": {
+          "env": "BACKEND_PORT"
+        }
+      }
+    }
+  },
+  "requiredExtensions": {
+    "dependencies": "cnab.io/specs/depenencies.schema.json"
```
glyn: Needs to be an array.

jeremyrickard: Doh, forgot to save the example when I pushed the rest;. Just fixed that.

101-bundle-json.md
```diff
@@ -852,13 +852,131 @@ The fields are defined as follows:
   - `EXTENSION NAME`: a unique name for an extension. Names SHOULD follow the dotted name format described earlier in this section.
   - The value of the extension must be valid JSON, but is otherwise undefined.
 
-The usage of extensions is undefined. However, bundles SHOULD be installable by runtimes that do not understand the extensions.
+### Required Extensions
+
+Some extensions MAY be required by a bundle and operations involving that bundle MUST fail if the runtime does not support that bundle. In this case, a bundle author MUST use the `requiredExtensions` array to define those extensions that are necessary for installing the bundle. The `requiredExtensions` array should contain the `EXTENSION NAME` defined in the `custom` object for any extensions that are required for bundle operations.
```
squillace: I'm with pedantry: "...are necessary for performing all actions with the bundle...." or some such.

101-bundle-json.md
```diff
@@ -852,13 +852,131 @@ The fields are defined as follows:
...
+
+A runtime SHOULD check that it supports any required custom actions before performing any operation. However, bundles SHOULD be installable by runtimes that do not support extensions that are NOT identified in the `requiredExtensions` field.
```
squillace: Have we decided on SHOULD? For extensions, SHOULD implies that a tool could not bother to check and end up running malbehaved invocationImages -- in the sense that they were intended to be run along with a specific extension that the runtime does not support. This, I dislike, in that there's no statement here that would create a requirement to indicate to the user that it cannot support the bundle. 

If I am reading this correctly, I would urge MUST on our tools here. Even non-supporting tools would then give the user an opportunity to do something different. With SHOULD, we seem to be saying `meh`. Do I misunderstand?

squillace: I'd very much like a statement SHOULD about the failure case for tooling. If a run time does not support any specified extension -- or extensions in general, should those different cases be called out? -- I'd very much like to emphasize that the proper way to handle this is to return as much information about the extensions not supported or about the extension feature itself so that users can have something to deal with as a result of the failure. This may be obvious, but asserting a SHOULD is a great emphasis for tool builders here.

101-bundle-json.md
squillace: My bad, @glyn, related to my comment https://github.com/deislabs/cnab-spec/pull/220#discussion_r302651720. I intended in my comment that MUST should be applied to the validation step, but NOT to the `force` activity step. For example, I would suggest that a runtime MUST inform the user that it doesn't know how to perform the custom extension X, but if the user passes a `--force` argument that it will perform the action in any case because "the user knows best". 

If that makes sense to us all, how should we state this? Maybe something like, "A runtime MUST check that it supports any required custom action... and it MUST require an explicit override statement if the user insists on attempting to perform an action on the bundle." I'm open to wording here, as my goal is to break apart the MUST CHECK AND STOP portion from the SHOULD ENABLE OVERRIDE portion.

101-bundle-json.md
jlegrone:
```suggestion
Some extensions MAY be required by a bundle and operations involving that bundle MUST fail if the runtime does not support that extension. In this case, a bundle author MUST use the `requiredExtensions` array to define those extensions that are necessary for installing the bundle. The `requiredExtensions` array should contain the `EXTENSION NAME` defined in the `custom` object for any extensions that are required for bundle operations.
```

jeremyrickard:
New proposal, based on input from @squillace up above.

```
A runtime MUST check that it supports any required custom actions before performing any operation. If the runtime does not support the required extension(s), it may proceed with the operation or fail, however it MUST notify the user that it does not support the required extensions. Bundles SHOULD be installable by runtimes that do not support extensions that are NOT identified in the `requiredExtensions` field.
```

@glyn is this too strongly worded? I think it allows runtimes to run...they just need to inform the user that they don't support a thing and it might not go well?

glyn:
> New proposal, based on input from @squillace up above.
> 
> ```
> A runtime MUST check that it supports any required custom actions before performing any operation. If the runtime does not support the required extension(s), it may proceed with the operation or fail, however it MUST notify the user that it does not support the required extensions. Bundles SHOULD be installable by runtimes that do not support extensions that are NOT identified in the `requiredExtensions` field.
> ```
> 
> @glyn is this too strongly worded? I think it allows runtimes to run...they just need to inform the user that they don't support a thing and it might not go well?

No, that looks ok to me. It's then up to a runtime how they structure this in terms of failure and "force" options.

101-bundle-json.md
glyn: I agree with @squillace here, but I wonder why the text switches from talking about required extensions to required custom actions. How about the following based on @jeremyrickard's suggested text?
```suggestion
A runtime MUST check that it supports any required extensions before performing any operation. If the runtime does not support the required extension(s), it MAY proceed with the operation or fail, however it MUST notify the user that it does not support the required extension(s). Bundles SHOULD be installable by runtimes that do not support extensions that are NOT identified in the `requiredExtensions` field. 
```

101-bundle-json.md
```diff
@@ -852,13 +852,131 @@ The fields are defined as follows:
   - `EXTENSION NAME`: a unique name for an extension. Names SHOULD follow the dotted name format described earlier in this section.
   - The value of the extension must be valid JSON, but is otherwise undefined.
 
-The usage of extensions is undefined. However, bundles SHOULD be installable by runtimes that do not understand the extensions.
+### Required Extensions
+
+Some extensions MAY be required by a bundle and operations involving that bundle MUST fail if the runtime does not support that extension. In this case, a bundle author MUST use the `requiredExtensions` array to define those extensions that are necessary for installing the bundle. The `requiredExtensions` array should contain the `EXTENSION NAME` defined in the `custom` object for any extensions that are required for bundle operations.
```
glyn: Please can be get rid of the confusing "In this case" and remove the bias towards installation thus:
```suggestion
Some extensions MAY be required in order for a runtime to any action on the bundle. A bundle author MUST use the `requiredExtensions` array to define those extensions that are required. The `requiredExtensions` array SHOULD contain the `EXTENSION NAME` defined in the `custom` object for each extension that is required.
```
```suggestion
Some extensions MAY be required by a bundle and operations involving that bundle MUST fail if the runtime does not support that extension. In this case, a bundle author MUST use the `requiredExtensions` array to define those extensions that are necessary for installing the bundle. The `requiredExtensions` array should contain the `EXTENSION NAME` defined in the `custom` object for any extensions that are required for bundle operations.
```

jeremyrickard:
@glyn incorporated your suggestions./ If you're happy with the state, feel free to LGTM2 and squash/merge!

glyn:
Apologies for failing to squash the commits for this PR.

PR #227, opened 2019-07-18
Make "required" a property of parameter

### LINKED ISSUE #222: Make specifying required parameters consistent with required credentials
Specifying required credentials is currently done via a boolean `required` property on each credential:

https://github.com/deislabs/cnab-spec/blob/8ec6c23db9bbc404be551ef13872c19fe03cc437/schema/bundle.schema.json#L22-L26

While required parameters are currently listed in a `required` array of strings referencing parameter keys:

```json
{
  "parameters": {
    "fields": {
      "backend_port": {
        "description": "The port that the backend will listen on",
        "destination": {
          "env": "PORT"
        },
        "immutable": true
    },
    "required": [
      "backend_port"
    ]
  }
}
```

This is also not internally consistent with how `immutable` is treated as a property of each parameter.

I propose that we refactor the parameter spec to remove `fields` and make `required` a property of the parameter type, eg:

```json
{
  "parameters": {
    "backend_port": {
      "description": "The port that the backend will listen on",
      "destination": {
        "env": "PORT"
      },
      "immutable": true,
      "required": true
    }
  }
}
```

jlegrone:
- a75deff34324bceb3883a1373f6e148df1d93167 Autoformats json with prettier (this was pretty helpful for fixing indentation everywhere)
- f599dcb00350238012ac721b9d02c8e3d42f5e40 Removes the `fields` and `required` properties from `parameters` in favor of `required` property directly on the parameter type.

Closes #222

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -115,13 +115,11 @@ The following is an example of a `bundle.json` for a bundled distributed as a _t
     }
   },
   "parameters": {
-    "fields": {
-      "backend_port": {
-        "definition": "http_port",
-        "description": "The port that the back-end will listen on",
-        "destination": {
-          "env": "BACKEND_PORT"
-        }
+    "backend_port": {
+      "definition": "http_port",
+      "description": "The port that the back-end will listen on",
+      "destination": {
+        "env": "BACKEND_PORT"
       }
     }
   },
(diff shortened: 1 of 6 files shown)


ryanmoran:
I do not oppose this. I don't have a strong preference for either this or the existing format. Both of them have trade offs and I think I'm more stuck on how we still have some parts of JSONSchema leaking out into the CNAB parameters spec.

The existing spec structure predates the addition of the `immutable` field which now makes that structure feel more objectionable. If we have two fields for specifying binary behaviors, it makes sense to align them to one another. But that means that this specification of `required` no longer aligns with how JSONSchema defines the `required` field.

For me, I don't feel like having two different ways of saying "this is required" is that bad, but I find that I'm usually that person whose brain latches onto mundane and arbitrary differences in technical details like this. For others, it might be a bit frustrating to have to continually discover this. Now, that statement is completely hypothetical at this point. We have roughly zero users for the spec and so have yet to really gather any sort of meaningful feedback besides our own.

---
On a related, but different note...
I find the recent addition of `immutable` both makes this change seem like the right step forward, and also introduces a set of features that overlaps with what we might already be able to achieve with JSONSchema alone. I'm not sure that in the long run we really want what `immutable` as a bespoke feature. Let me explain.

The behavior of `immutable` is that it allows a parameter to be written iff the bundle has not been installed. This behavior overlaps with what is already available in JSONSchema via the `readOnly` field. What differs here is that we cannot change the definition of a parameter based on what action is being run. This means that I cannot say that a field is `readOnly` for all actions other than `install`. However, if we did come up with some way of generically overriding parameters on a per-action basis, then you could implement `immutable` using existing primitives. Using a hypothetical 2.0 version of CNAB, I might be able to say the follow:

```json
{
  "definitions": {
    "writable_port": {
      "type": "integer",
    },
    "readonly_port": {
      "type": "integer",
      "readOnly": true
    }
  },
  "parameters": {
    "http_port": {
      "definition": "readonly_port",
      "destination": {
        "env": "GREETING"
      }
    }
  },
  "actions": {
    "install": {
      "parameters": {
        "http_port": {
          "definition": "writable_port",
          "destination": {
            "env": "GREETING"
          }
        }
      }
    }
  }
}
```
In the example above, I am able to define the `http_port` parameter as `readOnly`, but then override it for the `install` action to get the expected `immutable` behavior. This also makes this sort of per-action behavior modification much more flexible. I could optionally override the immutability of this field again in some `super-dangerous-upgrade` action at a later point without needing to enable that kind of feature in any CNAB runtime.

Due to how late in the 1.0 cycle the `immutable` feature was introduced, and the changes required to implement something like what is proposed here, I chose not to object to that addition. However, I think that it would be worth looking at the concept of `immutable` again in the future to see how we might be able to remove it, and therefore do less.

PR #228, opened 2019-07-18
Clarify why bundle.json is mounted in the invocation image

trishankatdatadog:
Signed-off-by: Trishank K Kuppusamy <trishank.kuppusamy@datadoghq.com>

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -26,7 +26,7 @@ A `bundle.json` is broken down into the following categories of information:
 - A list of outputs (name, type and location) that the application produces
 - A set of schema definitions used to validate user input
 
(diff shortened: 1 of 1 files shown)


101-bundle-json.md
```diff
@@ -26,7 +26,7 @@ A `bundle.json` is broken down into the following categories of information:
 - A list of outputs (name, type and location) that the application produces
 - A set of schema definitions used to validate user input
 
-The `bundle.json` is located at `/cnab/bundle.json` in the invocation image's runtime filesystem.
+The bundle definition is [made accessible](103-bundle-runtime.md#bundle-definition) from inside the invocation image in order to allow the run tool to reference information in the file. The bundle.json MUST be mounted to /cnab/bundle.json.
```
youreddy: @trishankatdatadog thanks for updating this section. The link is redundant since the same information is present in both places. What are your thoughts on either dropping the link or keeping the original wording and adding the link?

trishankatdatadog: I can remove the link

PR #230, opened 2019-07-22
Make contentDigest required in image JSON schema

trishankatdatadog:
Right now, it's ambiguous where `contentDigest` should be required or not in an image. This PR makes it required so that we can depend upon it for [CNAB Security](https://github.com/deislabs/cnab-spec/blob/master/300-CNAB-security.md).

### DIFF
--- a/schema/bundle.schema.json
+++ b/schema/bundle.schema.json
@@ -61,7 +61,7 @@
           "type": "integer"
         }
       },
-      "required": ["image"],
+      "required": ["image", "contentDigest"],
       "type": "object"
     },
     "invocationImage": {


trishankatdatadog:
Also, should I make this clear in 101?

glyn:
I like this change, but I wonder if `contentDigest` is applicable to _all_ image types. Hopefully so, but maybe @GreenCee or others would care to confirm.

trishankatdatadog:
What are all the image types, out of curiosity? There's containers, VMs... what else am I missing?

glyn:
> What are all the image types, out of curiosity? There's containers, VMs... what else am I missing?

I think it's open-ended, but OCI/docker images and VMs are the image types that have been discussed most.

trishankatdatadog:
Got it. I suspect that any image type should be hashable :)

jeremyrickard:
@simonferquel @silvin-lubecki can you weigh in on this from the Docker side?

jeremyrickard:
@trishankatdatadog based on the narrative text in the 101 markdown: 

```
The contentDigest field MUST contain a digest, in OCI format, to be used to compute the integrity of the image. The calculation of how the image matches the contentDigest is dependent upon image type. (OCI, for example, uses a Merkle tree while VM images are checksums.) If this field is omitted, a runtime is not obligated to validate the image. During bundle development, it may be ideal to omit the contentDigest field and/or skip validation. Once a bundle is ready to be transmitted as a thick or thin bundle, it must have a contentDigest field.
```

This wording implies that the field is optional during development, but it *must* have a `contentDigest` field when it is ready to be transmitted. I think your schema change implies that it must always be present. Is it sufficient for the security spec to state that it is in fact, required, for the security spec and that tooling that implements the security spec must report an error or something when the `contentDigest` is not present?

technosophos:
@jeremyrickard is correct. The reason this field is not marked required is so that during the development cycle we can run schema checks. Our intention was to not generate the hash until the packaging phase of the cycle. Generating a schema validation failure here felt awkward, since we'd basically be encouraging developers to add fake hashes during dev in order to get schema validation to pass.

The CNAB runtime, though, is still required to verify the presence of this field before trusting an image.

trishankatdatadog:
I see. OK, as long as the spec is clear that every `image` MUST have a `contentDigest`, we should be good to go!

PR #233, opened 2019-07-24
Remove the `immutable` attribute from parameters

### LINKED ISSUE #229: Resolve the `immutable` and `readOnly` behavior of parameters
For more descriptive text, see https://github.com/deislabs/cnab-spec/pull/227#issuecomment-513835322.

TL;DR, `readOnly` in the definition of a param seems to cover what `immutable` on a parameter is intended for. Is it needed as a bespoke functionality? 

Can we remove `immutable` and rely upon the `readOnly` attribute of the definition for now? Or do we need this as a separate thing (i.e. two params share the same definition but one is immutable and one is not?). I think it would be useful to try and figure out if we want to resolve this prior to 1.0 or if it's something we think we can deprecate in the future?

jeremyrickard:
This removes the `immutable` attribute from parameters per Issue #229

Closes #229

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -227,8 +227,7 @@ What follows is an example of a thick bundle. Notice how the `invocationImage` a
       "description": "The port that the backend will listen on",
       "destination": {
         "path": "/path/to/backend_port"
-      },
-      "immutable": true
+      }
     }
   },
   "schemaVersion": "v1.0.0-WD",
@@ -456,7 +455,6 @@ Parameter specifications consist of name/value pairs. The name is fixed, but the
     - `destination`: Indicates where (in the invocation image) the parameter is to be written (REQUIRED)
       - `env`: The name of an environment variable
       - `path`: The fully qualified path to a file that will be created. Specified path MUST NOT be a subpath of `/cnab/app/outputs`.
-    - `immutable`: an immutable parameter may only be set at installation. The value of the parameter cannot be changed. MUST be a boolean. Default value is false. (OPTIONAL).
     - `required`: indicates whether this parameter MUST be supplied. By default it is `false`, which means the parameter is optional. When `true`, a runtime MUST fail if the parameter is not provided for any action to which the parameter applies.
 
 Parameter names (the keys in `parameters`) ought to conform to the [Open Group Base Specification Issue 6, Section 8.1, paragraph 4](http://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap08.html) definition of environment variable names with one exception: parameter names MAY begin with a digit (approximately `[A-Z0-9_]+`).
@@ -536,7 +534,6 @@ The structure of a `parameters` and `definitions` section looks like the section
         "env": <string>,
         "path": <string>
       },
-      "immutable" : <boolean>,
       "required" : <boolean>
     }
   }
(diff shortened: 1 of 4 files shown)


radu-matei:
https://github.com/deislabs/cnab-spec/blob/17a57d809b82bd2ff178469cec79d1b76331f460/103-bundle-runtime.md#setting-parameter-values

> If an immutable parameter value is specified at any time after _install_, the CNAB Runtime MUST produce an error and discontinue action.

Should this also be reworded to reflect the removal of the `immutable` property?

jeremyrickard:
> https://github.com/deislabs/cnab-spec/blob/17a57d809b82bd2ff178469cec79d1b76331f460/103-bundle-runtime.md#setting-parameter-values
> 
> > If an immutable parameter value is specified at any time after _install_, the CNAB Runtime MUST produce an error and discontinue action.
> 
> Should this also be reworded to reflect the removal of the `immutable` property?

Yeah it should! Thought I got that reference. Thanks for catching it.

silvin-lubecki:
@jeremyrickard in the #229 issue you say:
> Can we remove immutable and rely upon the readOnly attribute of the definition for now?

Shouldn't we then add a hint in the specs that the immutability can be achieved using `readOnly` instead of removing every pieces of immutability reference? 🤔

jeremyrickard:
@silvin-lubecki maybe:

```
If an parameter contains the `readOnly` attribute and a value is specified at any time after install, the CNAB Runtime MUST produce an error and discontinue action.
```

@ryanmoran what do you think? `readOnly` says:

```
readOnly: Indicates that the value of the parameter cannot be modified. MUST be a boolean. (OPTIONAL)
```

Or is this really saying that w/in the context of an executing it can't be modified? We never get specific enough, and I think that's why the `immutable` attribute was introduced originally.  The `writeOnly` attribute is equally vague in my opinion. Once the parameter's value is set (i.e. we set it at runtime) what does that mean during execution? I think we should maybe spell out how these behave with both `parameters` and `outputs`.

astrieanna:
I'm concerned that this means you must run install first. What if a bundle has some other custom action that gets run first?

I think that `readOnly` and `writeOnly` apply to how individual actions can interact with specific parameters.

jeremyrickard:
@astrieanna yeah, I'm concerned that the behavior of those isn't well specified in the parameters section. The `writeOnly` especially to me, seems strange and likely to lead to confusion. How is that expressed to end users and what's the interaction between the runtime and the invocation image? It gets set by the runtime (written) and then can never be "read"). I think @ryanmoran was suggesting that in a later iteration of this, we could address different values for different actions to clearly state something like `readOnly for every action except for Install (since that's where you want it to be set`. 

I think I've convinced myself that I'm not sure that what we have is sufficient to express the immutability of something _today_ without the _immutable_ flag?  So either we drop `immutable` for 1.0, and think about how to introduce that post 1.0, or we should revisit this in a more in-depth call or something. 

@technosophos suggested just removing it in https://github.com/deislabs/cnab-spec/issues/229#issuecomment-514305482 and I think I'm inclined to do that and just remove all mentions of immutability for now (cc: @silvin-lubecki)

youreddy:
@jeremyrickard thanks for taking on this PR! I'm cool with dropping immutable for 1.0 and talking about it further. 

I was taking a look at the JSON Schema documentation today and something interesting I learned was that `readOnly` and `writeOnly` are annotations and not necessarily validations that cause an error. According to the [documentation](https://json-schema.org/latest/json-schema-validation.html#rfc.section.10.3) changing a `readOnly` property should be either ignored or rejected by the runtime. So yeah, I think we need to think more about how to use these properties to achieve the same outcomes as `immutable`. Especially if validation errors are a desired outcome.

jeremyrickard:
@glyn yes, I agree. I will open an issue and copy all of the dicussion into that so we can preserve it and revisit.

PR #278, opened 2019-09-09
Bump bundle schemaVersion to v1.0.0

chris-crone:
As part of the Core 1.0 release, we did not bump the bundle schemaVersion from v1.0.0-WD to v1.0.0.

cc @rumpl who caught this.


technosophos:
In the future, this should be noted in an errata section

PR #253, opened 2019-08-15
CNAB Security 300

trishankatdatadog:
Major TODOs:
- [ ] Add a threat model to basically guide what 
- [x] Talk about signing only bundles, not images, so simplify security levels
- [ ] Discuss use cases to talk about what should be possible (e.g., we are doing this to support that use case)
- [x] Explain that it doesn't matter where you store bundles or metadata, or whether it's a think or thick bundle, because verification would exactly the same

Minor TODOs:
- [x] Replace links to ITEs

Sub-PRs:
- [x] [301 to discuss what metadata repositories may look like](https://github.com/deislabs/cnab-spec/pull/281)
- [x] [302 to discuss prescriptive signing workflows](https://github.com/deislabs/cnab-spec/pull/282)
- [x] [303 to discuss mandatory verification workflows](https://github.com/deislabs/cnab-spec/pull/280)
- [x] [304 to discuss known implementations](https://github.com/deislabs/cnab-spec/pull/288)

PR #281, opened 2019-09-16
CNAB Security 301: Metadata repositories

trishankatdatadog:
Part of #253 

WIP: please do not merge yet!

Major TODOs:
- [x] Be more explicit about metadata living on OCI artifacts (or not, completely optional)
- [x] Start with concepts, and one good way to do it, for example everything with online keys first
- [x] We can talk about the four top-level roles, distinguish between online and offline keys, talk about possible delegations (keep it optional), talk about one possible implementation for one project
- [x] Radu: will work on known implementations
- [x] Move multiple projects to an Appendix, it's too complicated
- [x] Security analysis


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
radu-matei: (I usually add these to my global dictionary, not sure they should be in the repo).

301-metadata-repositories.md
```diff
@@ -0,0 +1,307 @@
...
+
+This document is a _normative_ part of [CNAB Security](300-CNAB-security.md).
+
+The keywords MUST, MUST NOT, REQUIRED, SHALL, SHALL NOT, SHOULD, SHOULD NOT, RECOMMENDED, MAY, and OPTIONAL in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119). The use of these keywords in this document are statements about how a CNAB implementation may fulfill the CNAB Security specification _only_.
+
+## Metadata repositories
+
+A _metadata repository_ is a service that hosts TUF and / or in-toto metadata about bundles and / or images. The repository is conceptually distinct from a [CNAB registry](200-CNAB-registries.md), which hosts bundles and / or images themselves.
+
+How TUF and in-toto metadata should be designed for a metadata repository depends on which purpose it serves. The repository may be _private_ for internal consumption, or _public_. Authentication to private metadata repositories are out of the scope of this document.
```
radu-matei: However, should the document point to some helper document detailing TUF authentication? Do we have that anywhere?

301-metadata-repositories.md
```diff
@@ -0,0 +1,307 @@
...
+
+This subsection discusses how an organization (e.g., `example.com/example-org/*`) MAY setup a metadata repository that hosts metadata about bundles and / or images developed and / or maintained by different _projects_, groups of developers. This repository MAY be private or public.
+
+![Figure 1: The suggested metadata repository for bundles and / or images developed and / or maintained by different projects, or groups of developers](img/example-metadata-repository.png)
+
+Figure 1 illustrates this suggested metadata repository, which we discuss using a top-down, outside-in approach.
+
+#### Administrators
+
+The four-top level TUF `root`, `timestamp`, `snapshot`, and `targets` roles SHOULD be controlled by the repository _administrators_, who develop and maintain the repository's hardware and software.
```
radu-matei: I would link all TUF and in-toto terms to the relevant sections in the respective specifications.

301-metadata-repositories.md
```diff
@@ -0,0 +1,307 @@
...
+
+The `timestamp` role indicates concisely whether there is any new metadata, bundle, or image on the repository.
+
+The `snapshot` role prevents attackers from tampering with interdependencies between bundles and / or images.
+
+Since metadata for both the `timestamp` and `snapshot` roles could be updated whenever a new bundle/ image is produced at any time, they MAY each use a threshold (1, 1) of _online keys_, or signing keys accessible on-demand by automation for the repository. In fact, they MAY even share the same online key. The metadata for each role MAY expire daily, since it is expected to be updated relatively frequently (e.g., whenever a developer pushes a new bundle / image, or a scheduled job runs to renew the metadata).
```
radu-matei: nit: space before _/_ in _bundle/ image_.

301-metadata-repositories.md
```diff
@@ -0,0 +1,307 @@
...
+
+Instead of signing bundles and images itself, the `targets` role SHOULD _delegate_ bundles and images to the `bundles` and `images` roles respectively. With a delegation, a _delegator_ binds a threshold of public keys, which belongs to a _delegatee_, to a _path pattern_, which specifies bundles or images (and in-toto metadata, if any) that the delegatee may be trusted with if the delegator has not signed for them.
+
+More concretely, the `targets` role SHOULD make two backtracking delegations. First, it should delegate all bundles (e.g., `['example.com/example-org/*:*', 'bundles/*']`) to the `bundles` role. Second, it SHOULD delegate all images (e.g., `['example.com/example-org/*:*', 'images/*']`) to the `images` role. In the interest of space, we will discuss only bundles going forward, but note that exactly the same principles apply to images.
```
radu-matei: I like this section, and I think it is really important to have it in this document - but would it make more sense to first show what it would mean to have the `targets` role directly sign bundles and see the disadvantages?

Then, organizations that are comfortable with the drawbacks could choose to use it directly, and those who need the additional delegations would better understand why they need them.

301-metadata-repositories.md
```diff
@@ -0,0 +1,307 @@
...
+
+The `targets` role SHOULD use a threshold (m, n) of offline keys, where n is the number of administrators, and m is the number of quorum members that must agree on new `targets` metadata. The keys between the `targets`, `images`, and `bundles` roles MAY all be shared. Its metadata SHOULD expire yearly, considering that this information should rarely change.
```
radu-matei: This would be a good point to link to additional reading about offline key treshholds.

CNAB.code-workspace
```diff
@@ -0,0 +1,7 @@
+{
```
radu-matei: Same for this file, not sure we should add it to the repo.

301-metadata-repositories.md
trishankatdatadog: Authentication to a private metadata repository is strictly outside the scope of TUF.

301-metadata-repositories.md
trishankatdatadog: We could, but my concern is: how many key mgmt models should we discuss? The more we discuss, the longer the document, the less people are likely to read, the more confusing they will find it, and the more we will have to further subdivide pieces. Consider how this would add to the present proposed levels of gradual security:

Level 0: no TUF or in-toto.
Level 1a: TUF for bundles.
Level 1b: TUF for images.
Level 2: TUF for bundles and images.
Level 2b: TUF + in-toto for bundles. TUF for images.
Level 2c: TUF for bundles. TUF + in-toto for images.
Level 3: TUF + in-toto for bundles and images.

Presently, these levels do not account for whether critical signing keys are online or offline (which affects security), and that is highly specific for the key management model.

Don't get me wrong, I think there is a sweet spot somewhere, something like PEP [458](https://www.python.org/dev/peps/pep-0458/) vs [480](https://www.python.org/dev/peps/pep-0480/). We should agree on them before going further. What do you think?

301-metadata-repositories.md
trishankatdatadog: A difference in style, but I fixed it.

trishankatdatadog:
Unless @radu-matei has major objections, I am going to consider this PR done for now, and move on to 302 and 303

trishankatdatadog:
Sorry this took so long, but was busy with another TUF + in-toto project at work!

This is now ready for review. Could @radu-matei or @squillace take a look? Thanks!

301-metadata-repositories.md
```diff
@@ -0,0 +1,158 @@
...
+* Both TUF and in-toto are frameworks that can be configured in a wide variety of ways to achieve varying degrees of security. However, we discuss one [minimum viable product (MVP)](#minimum-viable-product-mvp), which has been optimized in particular for [Docker Content Trust / Notary 0.6.1](https://github.com/theupdateframework/notary/releases/tag/v0.6.1), in this document.
+* Although we discuss only the signing and verification of bundles, exactly the same principles apply to images.
+
+## Minimum viable product (MVP)
+
+This subsection discusses the simplest way that developers MAY set up a metadata repository for their bundle. Every bundle MAY use a separate metadata repository on the same server, even if two or more bundles are maintained by the same group of developers. (This is similar to how images are signed using [Docker Content Trust](https://docs.docker.com/engine/security/trust/content_trust/). As noted earlier, the MVP has been optimized for [Docker Content Trust / Notary 0.6.1](https://github.com/theupdateframework/notary/releases/tag/v0.6.1).) Figure 1 illustrates our simple metadata repository for a bundle.
```
glyn: > Every bundle MAY use a separate metadata repository on the same server, [...]

I'd like this document to describe the implications of using _multiple_ servers to host metadata repositories. A usecase is where a user consumes bundles from multiple vendors who do not share a single, central server.

trishankatdatadog: I can add a note here about how developers may setup different metadata repos on different servers. And then, in 303 (verification workflows), I'll talk about how a bundle runtime can pull different bundles from different metadata repos on different servers. Would this help?

glyn: I think so, especially if there are some non-obvious gotchas.

trishankatdatadog: Does the new note help, or would you like more details?

glyn: It helps a little, but don't think the document really describes the security implications of using multiple servers to host metadata repositories. For example, it would be good to spell out how to establish trust in a new metadata repository - presumably mainly by reputation and guess-work? You see, I think this is a weakness of this whole approach in that scenario and it would be good to call that out so that there are no surprises down the line.

301-metadata-repositories.md
```diff
@@ -0,0 +1,158 @@
...
+}
+```
+
+### Security analysis of MVP
+
+Table 1 presents a summary of possible attacks given the key compromise of one or more TUF roles and in-toto functionaries. We assume that attackers can:
+
+1. Compromise any combination of keys as per any row in Table 1.
+1. Serve new TUF and in-toto metadata as well as bundles. This can be done either with a man-in-the-middle (MitM) attack, or by compromising the metadata repository and CNAB registry.
+
```
technosophos: I'd suggest being very specific about what "yes" and "no" mean in the table below, e.g. `Columns marked "yes" indicate that an attack is theoretically possible`

trishankatdatadog:
Thanks very much, @technosophos!

@glyn Would you be happy if we merge this PR as a WD for now, and address your concern about clarifying multiple repositories in a follow-up PR?

@radu-matei Would you please 👀 and ✅?

glyn:
> @glyn Would you be happy if we merge this PR as a WD for now, and address your concern about clarifying multiple repositories in a follow-up PR?

Sure. I raised https://github.com/cnabio/cnab-spec/issues/314 so this doesn't get forgotten.

trishankatdatadog:
Forced-push to keep DCO happy 😒

trishankatdatadog:
@radu-matei Can you please 👀 and ✅ ?

trishankatdatadog:
@radu-matei Done? Can you please merge? :)

trishankatdatadog:
@radu-matei Done! Can you please merge?

PR #289, opened 2019-10-23
Update 103-bundle-runtime.md

ventsip:
fixing the misleading description in the example of setting parameter value in file.


vdice:
Thanks for the fix @ventsip !

PR #290, opened 2019-10-30
Make status.json a CNAB Output

jlegrone:
As per https://github.com/deislabs/cnab-go/pull/154/files#r340791888, it seems like there should be no need for a separate `io.cnab.status+json` action when `io.cnab.status` also exists given that `io.cnab.status` could define a cnab output for the `status.json` file.

### DIFF
--- a/804-well-known-custom-actions.md
+++ b/804-well-known-custom-actions.md
@@ -11,8 +11,7 @@ A CNAB indicates that it supports those actions by including them in its custom 
 - `io.cnab.dry-run` (with `stateless`: true and `modifies`: false): execute the installation in a dry-run mode, allowing to see what would happen with the given set of parameter values.
 - `io.cnab.help` (with `stateless`: true and `modifies`: false): print an help message to the standard output. Implementations MAY print different messages depending on the parameters values passed to the invocation image.
 - `io.cnab.log` (with `stateless`: false and `modifies`: false): print logs of the installed system to the standard output.
-- `io.cnab.status` (with `stateless`: false and `modifies`: false): print a human readable status message to the standard output.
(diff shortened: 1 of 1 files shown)


804-well-known-custom-actions.md
```diff
@@ -11,8 +11,7 @@ A CNAB indicates that it supports those actions by including them in its custom
 - `io.cnab.dry-run` (with `stateless`: true and `modifies`: false): execute the installation in a dry-run mode, allowing to see what would happen with the given set of parameter values.
 - `io.cnab.help` (with `stateless`: true and `modifies`: false): print an help message to the standard output. Implementations MAY print different messages depending on the parameters values passed to the invocation image.
 - `io.cnab.log` (with `stateless`: false and `modifies`: false): print logs of the installed system to the standard output.
-- `io.cnab.status` (with `stateless`: false and `modifies`: false): print a human readable status message to the standard output.
-- `io.cnab.status+json` (with `stateless`: false and `modifies`: false): print a json payload describing the detailed status as in the example below ([here](schema/status.schema.json) is the JSON schema for this):
```
silvin-lubecki: I know this is only well known actions and not normative, but removing `io.cnab.status+json` may be considered as a breaking change? Do we know any bundle already implementing it? Shouldn't we instead deprecate it before removing it?

jlegrone: Personally I'm in favor of removal since IMO this is a pattern we should specifically avoid.

I don't consider this as a breaking change since as you mentioned this is a non-normative section of the spec, and removing documentation on `io.cnab.status+json` as a "well-known" custom action does not make the action itself invalid, as implementations are still required by the spec to support arbitrary custom actions.

technosophos: Breaking changes are currently allowed in all sections >= 200, since only CNAB Core has left Working Draft state.

silvin-lubecki: Fine by me 👍

PR #295, opened 2019-11-14
fix: formalize the field definitions on the Claims spec.

technosophos:
I made an earnest effort to be more specific about the requirements of the various fields on the Claims object.

### DIFF
--- a/400-claims.md
+++ b/400-claims.md
@@ -18,22 +18,6 @@ The claims system is designed to satisfy the requirements of the [Bundle Runtime
 
 This specification uses the terminology defined in [the CNAB Core specification](100-CNAB.md).
 
-## Concepts of Package Management
-
-A _package_ is a discrete data chunk that can be moved from location to location, and can be unpacked and installed onto a system. Typically, a package contains an application or application description. All package managers provide some explicit definition of a package and a package format.
-
-When a package is installed, the contents of a package are extracted and placed into the appropriate spaces on the target system, thus becoming an _installation_ (or _instance_) of the package.
-
-There are three core feature categories of a package manager system:
-
-- It can _install_ packages (initially put something onto a system)
-- It can _query_ installations (to see what is installed)
-- It can _upgrade_ and _delete_ packages (in other words, it can perform additional mutations on an existing installation)
-
-Package managers provide a wealth of other features, but the above are standard across all package managers. (For example, most package managers also provide a way to query what packages are available for installation.)
-
-This specification explains how CNAB records are generated such that continuity can be established across applications. In other words, this describes how CNAB bundles can be treated analogously to traditional packages.
-
 ## Managing State
 
 Fundamentally, package managers provide a state management layer to keep records of what was installed. For example, [homebrew](http://homebrew.sh), a popular macOS package manager, stores records for all installed software in `/usr/local/Cellar`. Helm, the package manager for Kubernetes, stores state records in Kubernetes ConfigMaps located in the system namespace. The Debian Apt system stores state in `/var/run`. In all of these cases, the stored state allows the package managing system to be able to answer (quickly) the question of whether a given package is installed.
--- a/400-claims.md
+++ b/400-claims.md
@@ -74,46 +74,6 @@ The CNAB claim is defined as a JSON document. The specification currently does n
 ```
 Source: [400.01-claim.json](examples/400.01-claim.json)
 
-<<<<<<< HEAD
-- bundle: The bundle, as defined in [the Bundle Definition](101-bundle-json.md).
-- bundleReference (OPTIONAL): A canonical reference to the bundle used in the last action. This bundle reference SHOULD be digested to identify a specific version of the referenced bundle.
-- created: A timestamp indicating when this release claim was first created. This MUST not be changed after initial creation.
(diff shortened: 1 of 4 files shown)


400-claims.md
```diff
@@ -50,13 +50,11 @@ From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/cscope.rb
 
 The CNAB Claims specification does not define where or how records are stored, nor how these records MAY be used by CNAB tooling. However, this specification describes how a CNAB-based system MUST emit the record to an invocation image, and provides some guidance on maintaining integrity of the system.
 
-This is done so that implementors can standardize on a way of relating a release claim (the record of a release) to release operations like `install`, `upgrade`, or `uninstall`. This, in turn, is necessary if CNAB bundles are expected to be executable by different implementations.
+This is done so that implementors can standardize on a way of relating a release claim (the record of a release) to release actions like `install`, `upgrade`, or `uninstall`. This, in turn, is necessary if CNAB bundles are expected to be executable by different implementations. If CNAB runtimes implement this common claims system, then tools can share responsibility for managing CNAB applications. For example, one tool may perform actions, while another may provide a visualization dashboard.
```
vdice: My only question here, which admittedly isn't one of the parts amended, is the term `release` as used in _"... a release claim", "... the record of a release"_, etc.

Should we update to `installation` ?  Or are they indeed distinct terms?

vdice:
(Apologies, CI is still in transition (hence the stalled `validation` check).  Since this PR does not change any schema, we're fine to go without.)

400-claims.md
technosophos: I'm going to do a follow-up PR on that. I need to do a full pass through the doc for the terms "release" and "installation" and make sure we consistently use them. An "installation" is when a package is installed. A "release" is each state change on an "installation" -- e.g. I install version 1, then upgrade to version 2: that is one installation with two releases. But there had been some other suggestions about how we might clarify that.

400-claims.md
```diff
@@ -91,24 +89,26 @@ The CNAB claim is defined as a JSON document. The specification currently does n
...
-    - underway: in progress. This should only be used if the invocation container MUST exit before it can determine whether all operations are complete. Note that `underway` is a _long term status_ that indicates that the installation's final state cannot be determined by the system. For this reason, it should be avoided.
-    - unknown: the state is unknown. This is an error condition.
-    - success: completed successfully
-- revision: An [ULID](https://github.com/ulid/spec) that MUST change each time the release is modified.
-
-> Note that credential data is _never_ stored in a claim. For this reason, a claim is not considered _trivially repeatable_. Credentials MUST be supplied on each action.
+The fields above are defined as follows:
+
+- `bundle` (REQUIRED): The bundle, as defined in [the Bundle Definition](101-bundle-json.md).
+- `created` (OPTIONAL): A timestamp indicating when this release claim was first created. This MUST not be changed after initial creation.
```
jlegrone: It would be nice if created/modified were required. Any particular reasons for keeping them optional? Also should we specify the timestamp format here?

```suggestion
- `created` (REQUIRED): A timestamp indicating when this release claim was first created. This MUST not be changed after initial creation.
```

400-claims.md
```diff
@@ -91,24 +89,26 @@ The CNAB claim is defined as a JSON document. The specification currently does n
...
+- `custom` (OPTIONAL): A section for custom extension data applicable to a given runtime.
+- `modified` (OPTIONAL): A timestamp indicating the last time this release claim was modified. The installation operation MAY set this to a timestamp that matches `created`. Any operation that changes the release SHOULD set this field to the time of the operation.
+- `name` (REQUIRED): The name of the _installation_. This can be automatically generated, though humans may need to interact with it. It MUST be unique within the installation environment, though that constraint MUST be imposed externally. Elsewhere, this field is referenced as the _installation name_. The format of this field must follow the same format used for the `name` field in the [bundle.json file specification](101-bundle-json.md#the-bundlejson-file).
```
jlegrone: This has been slightly confusing for me on occasion. Perhaps renaming this property to `installation` would provide more context when working with claims?

```suggestion
- `installation` (REQUIRED): The name of the _installation_. This can be automatically generated, though humans may need to interact with it. It MUST be unique within the installation environment, though that constraint MUST be imposed externally. Elsewhere, this field is referenced as the _installation name_. The format of this field must follow the same format used for the `name` field in the [bundle.json file specification](101-bundle-json.md#the-bundlejson-file).
```

400-claims.md
technosophos: I don't remember why `created` was required. `modified` was required because it should not be set until a record is actually modified.

400-claims.md
technosophos: Can we open this as an issue and circulate it a little. I think @carolynvs may have some thoughts on this as well, but she is out on holiday.

technosophos: See #303. Since this is a breaking change, and nothing else in the present PR is breaking, I would like to do it separately.

vdice:
This still looks good to me, although appears to need a rebase as well as signed commits to make DCO happy.

technosophos:
Closing in favor of #309, since I screwed up the rebase pretty badly here.

PR #299, opened 2019-12-11
add bundleReference to the claim specification

### LINKED ISSUE #297: A claim should have a reference to the bundle
Currently when we read a claim, there is no way to know which bundle has been originally used, as we only store a copy of the bundle itself and not a reference to the original one. 
This means that on a later action on an installation (upgrade, custom action...), as we only have the bundle copy from the claim, we have several problems:
* we lose all the bundle signing benefits, so I guess someone can temper with the installation during the upgrade
* we don't know where is stored the relocation map (on Docker App we have a bundle store which stores the relocation map along with the bundle itsefl)
I propose we add a new optional field "bundle reference", now that the bundle reference is better known. That way we can even pinpoint the bundle ref using a digested reference.

technosophos:
This defines an optional bundle reference as described in #297. I am not sure whether we should also have a relocation map stored inside of the claim, but if so this should be a separate PR.

Closes #297

### DIFF
--- a/400-claims.md
+++ b/400-claims.md
@@ -59,6 +59,7 @@ The CNAB claim is defined as a JSON document. The specification currently does n
     "schemaVersion": "v1.0.0",
     "version": "0.1.0"
   },
+  "bundleReference": "hub.example.com/my/bundle@sha256:eeeeeeeee...",
   "created": "2018-08-30T20:39:55.549002887-06:00",
   "custom": {},
   "modified": "2018-08-30T20:39:59.611068556-06:00",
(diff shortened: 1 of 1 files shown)


400-claims.md
```diff
@@ -76,6 +77,7 @@ The CNAB claim is defined as a JSON document. The specification currently does n
 Source: [400.01-claim.json](examples/400.01-claim.json)
 
 - bundle: The bundle, as defined in [the Bundle Definition](101-bundle-json.md).
+- bundleReference (OPTIONAL): A canonical reference to the bundle used in the last action. This bundle SHOULD use a SHA to identify the specific version of the referenced bundle.
```
glyn: Is this an OCI-style reference to a bundle stored in a repository or is it looser than that?

PR #305, opened 2020-01-07
Indicate that both parameters and credentials may inject files.

evankanderson:
In the `bundle.json` section, the `destination` of `parameters` may point to `env` or `path`. I'm guessing this text was left from an earlier version of the spec.

### DIFF
--- a/103-bundle-runtime.md
+++ b/103-bundle-runtime.md
@@ -30,7 +30,7 @@ The bundle definition is made accessible from inside the invocation image in ord
 CNAB allows injecting data into the invocation image in two ways:
 
 - Environment Variables: This is the preferred method. In this method, data is encoded as a string and placed into the the environment with an associated name.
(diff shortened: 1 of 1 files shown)


silvin-lubecki:
Hello @evankanderson , thank you for your contribution 👍
The `DCO bot` looks angry, can you sign off your commit to make him happy? 😄

technosophos:
To fix this, you just need to use `--signoff`. I will update the local docs for this, but in the meantime this guide should get you started: https://kauri.io/dco-sign-off:-commiting-code-to-hyperledger-besu/f58190e5e3bc4b1a9ed902bfccfe58b9/a

evankanderson:
I fixed the DCO; I tried a new github web UI flow and it turns out that it doesn't including the --signoff when it creates the branch and performs the diffs.

PR #309, opened 2020-01-09
Formalize Claims fields

technosophos:
This supersedes #295 

This PR fixes the field definitions on the claims spec to be more specific.

### DIFF
--- a/400-claims.md
+++ b/400-claims.md
@@ -82,33 +82,12 @@ The CNAB claim is defined as a JSON document. The specification currently does n
 ```
 Source: [400.01-claim.json](examples/400.01-claim.json)
 
-<<<<<<< HEAD
-- bundle: The bundle, as defined in [the Bundle Definition](101-bundle-json.md).
-- bundleReference (OPTIONAL): A canonical reference to the bundle used in the last action. This bundle reference SHOULD be digested to identify a specific version of the referenced bundle.
-- created: A timestamp indicating when this claim was first created. This MUST not be changed after initial creation.
-- custom: A section for custom extension data applicable to a given runtime.
-- modified: A timestamp indicating the last time this claim was modified. This MUST be updated each time the claim is modified.
-- name: The name of the _installation_. This can be automatically generated, though humans may need to interact with it. It MUST be unique within the installation environment, though that constraint MUST be imposed externally. Elsewhere, this field is referenced as the _installation name_.
-- outputs: Key/value pairs that were created by the operation. These are stored so that the user can access them after the operation completes. Some implementations MAY choose not to store these for security or portability reasons.
-- parameters: Key/value pairs that were passed in during the operation. These are stored so that the operation can be re-run. Some implementations MAY choose not to store these for security or portability reasons.
-- result: The outcome of the bundle's last action (e.g. if action is install, this indicates the outcome of the installation.). It is an object with the following fields:
-  - action: Indicates the action that the current bundle is in. This may be any of the built-in actions (`install`, `upgrade`, `uninstall`) as well as any custom actions as defined in the bundle descriptor. The special name `unknown` MAY be used in the case where the CNAB Runtime cannot determine the action name of a claim.
-  - message: A human-readable string that communicates the outcome. Error messages MAY be included in `failure` conditions.
-  - status: Indicates the status of the last phase transition. Valid statuses are:
-    - failure: failed before completion
-    - underway: in progress. This should only be used if the invocation container MUST exit before it can determine whether all operations are complete. Note that `underway` is a _long term status_ that indicates that the installation's final state cannot be determined by the system. For this reason, it should be avoided.
-    - unknown: the state is unknown. This is an error condition.
-    - success: completed successfully
-- revision: An [ULID](https://github.com/ulid/spec) that MUST change each time the installation is modified. It MUST NOT change when a [non-modifying operation](https://github.com/cnabio/cnab-spec/blob/master/101-bundle-json.md#custom-actions) is performed on an installation.
-
-> Note that credential data is _never_ stored in a claim. For this reason, a claim is not considered _trivially repeatable_. Credentials MUST be supplied on each action.
-||||||| parent of a55a1cd... Update 400-claims.md
 The fields above are defined as follows:
 
 - `bundle` (REQUIRED): The bundle, as defined in [the Bundle Definition](101-bundle-json.md).
 - `created` (OPTIONAL): A timestamp indicating when this release claim was first created. This MUST not be changed after initial creation.
 - `custom` (OPTIONAL): A section for custom extension data applicable to a given runtime.
-- `modified` (OPTIONAL): A timestamp indicating the last time this release claim was modified. The installation operation MAY set this to a timestamp that matches `created`. Any operation that changes the release SHOULD set this field to the time of the operation.
+- `modified` (OPTIONAL): A timestamp indicating the last time this release claim was modified. The installation operation MAY set this to a timestamp that matches `created`. Any operation that changes the claim MUST set this field to the time of the operation.
 - `name` (REQUIRED): The name of the _installation_. This can be automatically generated, though humans may need to interact with it. It MUST be unique within the installation environment, though that constraint MUST be imposed externally. Elsewhere, this field is referenced as the _installation name_. The format of this field must follow the same format used for the `name` field in the [bundle.json file specification](101-bundle-json.md#the-bundlejson-file).
 - `outputs` (OPTIONAL): Key/value pairs that were created by the operation. These are stored so that the user can access them after the operation completes. Some implementations MAY choose not to store these for security or portability reasons. See the [Outputs](#outputs) section for details. If this field is not present, it may be assumed that no outputs were generated as a result of the operation.
 - `parameters` (OPTIONAL): Key/value pairs that were passed in during the operation. These are stored so that the operation can be re-run. Some implementations MAY choose not to store these for security or portability reasons. However, there are some caveats. See the [Parameters](#parameters) section below for details.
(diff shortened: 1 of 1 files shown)


400-claims.md
```diff
@@ -76,25 +76,26 @@ The CNAB claim is defined as a JSON document. The specification currently does n
...
-    - failure: failed before completion
-    - underway: in progress. This should only be used if the invocation container MUST exit before it can determine whether all operations are complete. Note that `underway` is a _long term status_ that indicates that the installation's final state cannot be determined by the system. For this reason, it should be avoided.
-    - unknown: the state is unknown. This is an error condition.
-    - success: completed successfully
-- revision: An [ULID](https://github.com/ulid/spec) that MUST change each time the release is modified.
-
-> Note that credential data is _never_ stored in a claim. For this reason, a claim is not considered _trivially repeatable_. Credentials MUST be supplied on each action.
+The fields above are defined as follows:
+
+- `bundle` (REQUIRED): The bundle, as defined in [the Bundle Definition](101-bundle-json.md).
```
silvin-lubecki: It looks like `bundleReference (OPTIONAL)` is missing here ? 🤔

400-claims.md
```diff
@@ -76,25 +76,26 @@ The CNAB claim is defined as a JSON document. The specification currently does n
...
+- `created` (OPTIONAL): A timestamp indicating when this release claim was first created. This MUST not be changed after initial creation.
+- `custom` (OPTIONAL): A section for custom extension data applicable to a given runtime.
+- `modified` (OPTIONAL): A timestamp indicating the last time this release claim was modified. Executing the installation action MAY set this to a timestamp that matches `created`. Executing any modifying action (`upgrade`, `uninstall`, or a custom action with the property `"modifies": true`) MUST set this field to the time of the operation.
```
silvin-lubecki: Should we specify that the operation needs to be successful?

jeremyrickard: I think the `message` field right below this specifies that things can be failed though?

400-claims.md
```diff
@@ -76,25 +76,26 @@ The CNAB claim is defined as a JSON document. The specification currently does n
...
+- `result` (REQUIRED): The outcome of the bundle's last action (e.g. if action is install, this indicates the outcome of the installation.). It is an object with the following fields:
+  - `action` (REQUIRED): Indicates the action that the current bundle is in. This may be any of the built-in actions (`install`, `upgrade`, `uninstall`) as well as any custom actions as defined in the bundle descriptor. The special name `unknown` MAY be used in the case where the CNAB Runtime cannot determine the action name of a claim.
+  - `message` (OPTIONAL): A human-readable string that communicates the outcome. Error messages MAY be included in `failure` conditions.
+  - `status` (REQUIRED): Indicates the status of the last phase transition. Valid statuses are:
+    - `failure`: failed before completion
+    - `underway`: in progress. This should only be used if the invocation container MUST exit before it can determine whether all operations are complete. Note that `underway` is a _long term status_ that indicates that the installation's final state cannot be determined by the system. For this reason, it should be avoided.
```
evankanderson: I don't see (in "Calculating the Result") how this value would ever be set, assuming that exit code 0 == `success`, exit code !0 == `failure` and timeout == `unknown`.

It might be worth clarifying in the "Calculating the Result" section how the different status values are determined.

jlegrone: I'm a little confused by 

> Note that `underway` is a _long term status_ that indicates that the installation's final state cannot be determined by the system.

Is `underway` not meant to be used to indicate that a mutation is currently being performed on the installation? I think this would be useful (and have been using a `pending` status to indicate this in internal tooling).

400-claims.md
```diff
@@ -76,25 +76,26 @@ The CNAB claim is defined as a JSON document. The specification currently does n
...
+- `name` (REQUIRED): The name of the _installation_. This can be automatically generated, though humans may need to interact with it. It MUST be unique within the installation environment, though that constraint MUST be imposed externally. Elsewhere, this field is referenced as the _installation name_. The format of this field must follow the same format used for the `name` field in the [bundle.json file specification](101-bundle-json.md#the-bundlejson-file).
+- `outputs` (OPTIONAL): Key/value pairs that were created by the operation. These are stored so that the user can access them after the operation completes. Some implementations MAY choose not to store these for security or portability reasons. See the [Outputs](#outputs) section for details. If this field is not present, it may be assumed that no outputs were generated as a result of the operation.
```
jlegrone: It seems like these two statements may conflict

> Some implementations MAY choose not to store these for security or portability reasons.

> If this field is not present, it may be assumed that no outputs were generated as a result of the operation.

If an implementation does not add outputs to a claim, it doesn't necessarily mean that no outputs were generated by the invocation image; we may need a way to indicate this.

In our internal implementation we don't currently store outputs in claims, but plan to allow querying them on demand. This is because the size of outputs is unbounded, and could make it difficult to scale to keeping many claims in memory. We also want to prevent oversending on the network when an output is generated but never read.

I wonder if it makes sense to simply have a separate object for a claim result, that includes outputs? Or perhaps we could remove outputs from claims and this could just be left as non-normative for now (they could always be included in the `custom` map).

carolynvs: > I wonder if it makes sense to have a separate object for a claim result, that includes outputs?

I'd like for us to consider that further as well. Outputs can be quite large and the contract for storing claims is quite limited (name + bytes). I'm interested in storing state adjacent to the claim. A common mechanism for storing data next to a claim (outputs, state), instead of inside of it, would give us a lot more flexibility.

jlegrone: Runtimes may also have use cases for computing digests for claims, ie. to generate a content-addressable index, which would be easier done without embedded state.

400-claims.md
carolynvs: I had the same expectations as @jlegrone. If tools avoid setting `underway` everytime they run a bundle  then it doesn't really address my concern from https://github.com/cnabio/cnab-spec/issues/312

400-claims.md
chris-crone: Is there any reason that we wouldn't want this set on claim creation?

```suggestion
- `modified` (OPTIONAL): A timestamp indicating the last time this release claim was modified. Executing the installation action MUST set this to a timestamp that matches `created`. Executing any modifying action (`upgrade`, `uninstall`, or a custom action with the property `"modifies": true`) MUST set this field to the time of the operation.
```

technosophos: The timestamp is for the claim itself. So if a system writes a claim for a failed install, and the claim it has written is different from the previous claim, this field must set this.

That said... I don't know that systems will choose to store claims for failed releases

400-claims.md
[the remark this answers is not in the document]
technosophos: That text wouldn't belong here,though, would it? It would belong in the section about outputs.

technosophos: There are two issues:
1. Are the two statements conflicting?
2. Should we store the outputs elsewhere?

To #1, no, they are not conflicting. Both are `MAY`, and so it is up to the runtime to decide how to handle those. If a runtime has no access to the values of outputs, an implementor may reasonably conclude that there are no outputs and proceed. Likewise, a runtime may determine that some (or all) outputs are too sensitive to store, and hence decide not to store them. This is also okay. This does mean that it is possible that a runtime may not be able to do some things as follow up operations if it does not have access to some outputs, but that is okay. We do not need to require that they do.

Now consider the converse: If a runtime saw no outputs, should it conclude that something was wrong? I don't think that makes sense.

If a runtime determines that a field is sensitive, should it be forced to store it if its storage is not secure? I don't think that makes much sense either.

So the tradeoff we're shooting for is that it's better for a runtime to not have access to all of the fields, yet still be able to perform some operations than for a runtime to be forced to do something unwise.

2. I'm fine with a proposal to split outputs into another document/object, but I don't think that should belong on this PR or block this PR.

400-claims.md
```diff
@@ -76,25 +76,26 @@ The CNAB claim is defined as a JSON document. The specification currently does n
...
+- `parameters` (OPTIONAL): Key/value pairs that were passed in during the operation. These are stored so that the operation can be re-run. Some implementations MAY choose not to store these for security or portability reasons. However, there are some caveats. See the [Parameters](#parameters) section below for details.
+- `result` (REQUIRED): The outcome of the bundle's last action (e.g. if action is install, this indicates the outcome of the installation.). It is an object with the following fields:
+  - `action` (REQUIRED): Indicates the action that the current bundle is in. This may be any of the built-in actions (`install`, `upgrade`, `uninstall`) as well as any custom actions as defined in the bundle descriptor. The special name `unknown` MAY be used in the case where the CNAB Runtime cannot determine the action name of a claim.
```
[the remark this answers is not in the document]
technosophos: What change are you suggesting on this PR? As I said before, adding a new PR to separate the claims and the outputs is fine, but I am not sure I see what you are suggesting change on this PR.

400-claims.md
technosophos: I did not change the wording of `underway` between the previous version and this one. I'm just trying to get the formatting and all fields basically covered in this PR. A follow-up issue or something is fine with me, but we're holding up a formatting PR with unrelated things.

technosophos: Renaming to `pending` is up for discussion here: https://github.com/cnabio/cnab-spec/pull/325

technosophos: @carolynvs can you follow up on #312?

400-claims.md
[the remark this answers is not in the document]
technosophos: Opened a specific issue for this: https://github.com/cnabio/cnab-spec/issues/326

technosophos: Opened a new issue to discuss the outputs here: https://github.com/cnabio/cnab-spec/issues/327

400-claims.md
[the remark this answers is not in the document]
technosophos: See https://github.com/cnabio/cnab-spec/issues/329 and https://github.com/cnabio/cnab-spec/issues/328

technosophos:
I have opened issues for all remaining comments that were not simple fixes to this PR.

400-claims.md
carolynvs: Will do!

PR #311, opened 2020-01-09
Correct the wording for result.action

### LINKED ISSUE #251: When is a claim result updated?
The Claims spec (https://github.com/deislabs/cnab-spec/blob/master/400-claims.md) has this to say about the `result` attribute:  `result: The outcome of the bundle's last action`.  Additionally, the `result` contains a `status` which gives 'the status of the last phase transition' - this is an outcome such as `success` or `failure`.  Together these strongly imply that the result (and presumably other parts of the claim such as the `parameters` section) is updated when an action _completes_, i.e. the invocation container _exits_.  And indeed the spec says "The result object is populated by the result of the invocation image's action."

However, `result.action` is documented as "the action that the current bundle is in."  This strongly implies that if an action is in progress then it reflects _that_ action, not the last.

Is this just a wording glitch describing `result.action`?  And the claim is not updated at the beginning of an action, and does not indicate if an action is in progress?

technosophos:
This fixes the wording on `result.action`, which implied that the action was the _current_ action, not the _last completed_ action.

Depends upon #309 
Closes #251


silvin-lubecki:
Added `[On hold]` to the PR title until #309 is merged.

technosophos:
Rebased on 309

Issue #250, opened 2019-08-08
Spec does not specify how to stringise non-scalar parameter types


itowlson:
The spec allows parameters to be of type `object` and `array`.  But the runtime has to inject these parameter values into the invocation image via environment variables or files, which are inherently stringular.  There is no issue here for numbers, strings and booleans which have widely agreed`*` string representations; but we should state how runtimes should stringise object and array values, e.g. as JSON literals.

`*` Though not necessarily as widely agreed as you would like, e.g. a naive JavaScript implementation might easily inject numbers in scientific notation which, say, an invocation image binary written in Go might not readily parse.

astrieanna:
Kind of related, PR to fix Duffle injecting `object` and `array` as golang prints them: https://github.com/deislabs/cnab-go/pull/96

carolynvs:
Based on how the PR was resolved, do we need to update the spec to match what we did in cnab-go?

carolynvs:
Radu says that we don't have to worry about numbers because canonical json doesn't allow for anything but integers. 😀

carolynvs:
I'll update the spec to clarify that objects and arrays must be represented as json.

PR #320, opened 2020-02-12
replaced ambiguous description of environment variable content

### LINKED ISSUE #316: Spec is contradictory on how string-type parameter values are encoded
The [Setting Parameter Values section of the bundle runtime spec](https://github.com/cnabio/cnab-spec/blob/master/103-bundle-runtime.md#setting-parameter-values) seems to contradict itself. The parameter value evaluation rules state (#269) 
> Values are encoded as JSON strings.

But the examples in the section do not seem to follow this rule. The `path` destination example states that the `/var/run/greeting.txt` file would be created with contents `hello`, which is not well-formed JSON text. By my reading of the rules, the file should instead have the contents `"hello"`. Which is the correct behavior? Are the rules incomplete, or are the examples incorrect?

technosophos:
This addresses the root problems in #316, which were about encoding the data inside of env vars. Somehow, we gave the false impression that the body of an env var had to be JSON. I have backed out the text that seemed to imply that.

Note that I have gone with fairly minimal restraints (adopting Chris's SHOULD instead of a MUST) because we need to facilitate some interesting edge cases where a runtime needs to support uncommon character encodings or payload types.

Keep in mind that the author of the bundle.js is presumed to be an expert in the invocation image. So protecting the bundle author from the invocation image (and vice versa) is not a high priority. So this focuses mainly on the relationship between the runtime and the bundle. For a runtime that is tooled to support rare/bespoke operating systems (as we see in IoT), we opt for guidance that allows necessary deviation.

/cc @chris-crone and @corhere

Closes #316 

### DIFF
--- a/103-bundle-runtime.md
+++ b/103-bundle-runtime.md
@@ -140,11 +140,12 @@ The parameter value is evaluated thus:
 - If the CNAB runtime does not provide a value, but `default` is set, then the default value MUST be used.
 - If the parameter is marked `required` and `default` is set, then the requirement is satisfied by the runtime-provided default.
 - If no value is provided and `default` is unset, the runtime MUST set the value to an empty string (""), regardless of type.
+- When serializing non-string environment variable values, types SHOULD be converted to their [JSON equivalent type](https://www.json.org/json-en.html) and expressed in string format. For example, boolean `true` (or `True` or `TRUE`) SHOULD be expressed as `"true"` and the object `{foo: 23}` SHOULD be expressed as `"{\"foo\": 23}"`.
 - The resolved content of the environment variable SHOULD use UTF-8 character encoding.
 
 > Setting the value of other types to a default value based on type, e.g. Boolean to `false` or integer to `0`, is considered _incorrect behavior_. Setting the value to `null`, `nil`, or a related character string is also considered incorrect.
 
-In the case where the `destination` object has a `path` field, the CNAB runtime MUST create a file at that path. The file MUST have sufficient permissions that the effective user ID of the image can read the contents of the file. the file's character encoding SHOULD be UTF-8 and new lines SHOULD be LF (line feed). And the contents of the file MUST be the parameter value (calculated according to the rules above).
+In the case where the `destination` object has a `path` field, the CNAB runtime MUST create a file at that path. The file MUST have sufficient permissions that the effective user ID of the image can read the contents of the file. The file's character encoding SHOULD be UTF-8 and new lines SHOULD be LF (line feed). And the contents of the file MUST be the parameter value (calculated according to the rules above).
 
 ```json
 {
--- a/103-bundle-runtime.md
+++ b/103-bundle-runtime.md
@@ -140,7 +140,7 @@ The parameter value is evaluated thus:
 - If the CNAB runtime does not provide a value, but `default` is set, then the default value MUST be used.
 - If the parameter is marked `required` and `default` is set, then the requirement is satisfied by the runtime-provided default.
 - If no value is provided and `default` is unset, the runtime MUST set the value to an empty string (""), regardless of type.
-- When serializing non-string environment variable values, types SHOULD be converted to their [JSON equivalent type](https://www.json.org/json-en.html) and expressed in string format. For example, boolean `true` (or `True` or `TRUE`) SHOULD be expressed as `"true"` and the object `{foo: 23}` SHOULD be expressed as `"{\"foo\": 23}"`.
+- Non-string values SHOULD be converted to [JSON text](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf). For example, boolean `true` (or `True` or `TRUE`) SHOULD be expressed as `true` and the object `{foo: 23}` SHOULD be expressed as `{"foo":23}`. String values SHOULD NOT be converted to JSON text.
 - The resolved content of the environment variable SHOULD use UTF-8 character encoding.
 
 > Setting the value of other types to a default value based on type, e.g. Boolean to `false` or integer to `0`, is considered _incorrect behavior_. Setting the value to `null`, `nil`, or a related character string is also considered incorrect.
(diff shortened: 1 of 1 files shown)


103-bundle-runtime.md
```diff
@@ -140,11 +140,11 @@ The parameter value is evaluated thus:
 - If the CNAB runtime does not provide a value, but `default` is set, then the default value MUST be used.
 - If the parameter is marked `required` and `default` is set, then the requirement is satisfied by the runtime-provided default.
 - If no value is provided and `default` is unset, the runtime MUST set the value to an empty string (""), regardless of type.
-- Values are encoded as JSON strings.
+- The resolved content of the environment variable SHOULD be use UTF-8 character encoding.
```
chris-crone:
```suggestion
- The resolved content of the environment variable SHOULD use UTF-8 character encoding.
```

technosophos: Fixed! Thanks

103-bundle-runtime.md
```diff
@@ -140,11 +140,11 @@ The parameter value is evaluated thus:
...
-- Values are encoded as JSON strings.
+- The resolved content of the environment variable SHOULD use UTF-8 character encoding.
 
 > Setting the value of other types to a default value based on type, e.g. Boolean to `false` or integer to `0`, is considered _incorrect behavior_. Setting the value to `null`, `nil`, or a related character string is also considered incorrect.
 
-In the case where the `destination` object has a `path` field, the CNAB runtime MUST create a file at that path. The file MUST have sufficient permissions that the effective user ID of the image can read the contents of the file. And the contents of the file MUST be the parameter value (calculated according to the rules above).
+In the case where the `destination` object has a `path` field, the CNAB runtime MUST create a file at that path. The file MUST have sufficient permissions that the effective user ID of the image can read the contents of the file. the file's character encoding SHOULD be UTF-8 and new lines SHOULD be LF (line feed). And the contents of the file MUST be the parameter value (calculated according to the rules above).
```
chris-crone: One more minor issue that I missed on the first pass...

```suggestion
In the case where the `destination` object has a `path` field, the CNAB runtime MUST create a file at that path. The file MUST have sufficient permissions that the effective user ID of the image can read the contents of the file. The file's character encoding SHOULD be UTF-8 and new lines SHOULD be LF (line feed). And the contents of the file MUST be the parameter value (calculated according to the rules above).
```

corhere:
How should compliant runtimes serialize non-string parameters? See also #250.

technosophos:
Environment variables and files are always serialized to strings of bytes, no? I just defined the character encoding for that string. What else are you suggesting that I need to do? I suppose I could be explicit about strings.

technosophos:
Or are you asking what a runtime should encode a complex object into if someone passes a complex object instead of something that can easily be serialized into a string?

corhere:
> Or are you asking what a runtime should encode a complex object into if someone passes a complex object instead of something that can easily be serialized into a string?

Yes. And the other primitive types, too. Boolean values, for example, are easy to encode to strings in many languages, but the result is not standardized.

Compare Python's native encoding
```python
>>> "%s %s %s" % (True, False, None)
'True False None'
```
to Go's
```go
func main() {
	fmt.Printf("%v %v %v\n", true, false, nil)
	// Output: true false <nil>
}
```

technosophos:
Okay. Updated to provide what I think is the most consistent wording with what is currently in the spec.

103-bundle-runtime.md
```diff
@@ -140,11 +140,12 @@ The parameter value is evaluated thus:
 - If the CNAB runtime does not provide a value, but `default` is set, then the default value MUST be used.
 - If the parameter is marked `required` and `default` is set, then the requirement is satisfied by the runtime-provided default.
 - If no value is provided and `default` is unset, the runtime MUST set the value to an empty string (""), regardless of type.
-- Values are encoded as JSON strings.
+- When serializing non-string environment variable values, types SHOULD be converted to their [JSON equivalent type](https://www.json.org/json-en.html) and expressed in string format. For example, boolean `true` (or `True` or `TRUE`) SHOULD be expressed as `"true"` and the object `{foo: 23}` SHOULD be expressed as `"{\"foo\": 23}"`.
```
corhere:
```suggestion
- Non-string values SHOULD be converted to [JSON text](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf). For example, boolean `true` (or `True` or `TRUE`) SHOULD be expressed as `true` and the object `{foo: 23}` SHOULD be expressed as `{"foo":23}`. String values SHOULD NOT be converted to JSON text.
```

Environment variables and files are both unstructured byte-strings, so serialization is required for both. And I believe that my suggested wording more accurately describes cnab-go's existing behavior.

https://github.com/youreddy/cnab-go/blob/d3645483baf494770ddbc1d83545d20060dd8427/action/action_test.go#L205-L213

103-bundle-runtime.md
```diff
@@ -140,11 +140,12 @@ The parameter value is evaluated thus:
 - If the CNAB runtime does not provide a value, but `default` is set, then the default value MUST be used.
 - If the parameter is marked `required` and `default` is set, then the requirement is satisfied by the runtime-provided default.
 - If no value is provided and `default` is unset, the runtime MUST set the value to an empty string (""), regardless of type.
-- Values are encoded as JSON strings.
+- Non-string values SHOULD be converted to [JSON text](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf). For example, boolean `true` (or `True` or `TRUE`) SHOULD be expressed as `true` and the object `{foo: 23}` SHOULD be expressed as `{"foo":23}`. String values SHOULD NOT be converted to JSON text.
```
vdice: Perhaps this is just temporary/coincidence, but the link here is currently hanging for me when using the `http` protocol, as written.  When I switch to `https`, all is well: https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf

PR #321, opened 2020-02-12
removed unnecessary references to the original authors' registries

technosophos:
There were a few references to my DockerHub repo and Gabe's ACR repo. I replaced those with generic `example` repos.


technosophos:
Note that this is JUST the 1XX sections so that we can merge it only for the CNAB core spec. I will open a separate one for other specs

PR #323, opened 2020-02-12
add 'title' field for custom actions

### LINKED ISSUE #242: Support display names for actions
The CNAB WD says that action names "SHOULD use dotted name syntax".  This is great for uniqueness, but not great for UI.  Could we have an optional 'display name' attribute for actions?  The scenario is a GUI which displays the actions that can be taken on a bundle (or installation/instance/claim), e.g. as buttons or on a menu: it would be nice to be able to display:

```
--------------------------------------
|  Upgrade  |  Status  |  Uninstall  |
--------------------------------------
```

instead of

```
----------------------------------------------
|  Upgrade  |  io.cnab.status  |  Uninstall  |
----------------------------------------------
```

technosophos:
Adds a `display` field for custom actions so that a human friendly name can be shown for each custom action.

Closes #242 

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -749,7 +749,7 @@ Implementations MAY support user-defined additional actions as well. Such action
       "stateless": true
     },
     "io.cnab.migrate": {
-      "display": "Migratte",
+      "display": "Migrate",
       "modifies": false
     },
     "io.cnab.status": {
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -743,14 +743,17 @@ Implementations MAY support user-defined additional actions as well. Such action
 {
   "actions": {
     "io.cnab.dry-run": {
+      "display": "Dry Run",
       "description": "prints what install would do with the given parameters values",
       "modifies": false,
       "stateless": true
     },
     "io.cnab.migrate": {
+      "display": "Migratte",
       "modifies": false
     },
     "io.cnab.status": {
+      "display": "Status",
       "description": "retrieves the status of an installation",
       "modifies": false
     }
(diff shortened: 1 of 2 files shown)


101-bundle-json.md
```diff
@@ -743,14 +743,17 @@ Implementations MAY support user-defined additional actions as well. Such action
 {
   "actions": {
     "io.cnab.dry-run": {
+      "display": "Dry Run",
       "description": "prints what install would do with the given parameters values",
       "modifies": false,
       "stateless": true
     },
     "io.cnab.migrate": {
+      "display": "Migratte",
```
vdice: I believe this should be `"Magritte"` 😉 

art-magritte-listening-chamber-450x367px

schema/bundle.schema.json
```diff
@@ -187,6 +187,10 @@
             "description": "A description of the purpose of this action",
             "type": "string"
           },
+          "display": {
+            "description": "A human-readable name for thiis action",
```
vdice: Too many 👁 s in `thiis`

101-bundle-json.md
```diff
@@ -764,6 +767,7 @@ The above declares three actions: `io.cnab.status`, `io.cnab.migrate` and `io.cn
 
 Each action is accompanied by a description, which contains the following fields:
 
+- `display`: The human-readable name of the operation. If omitted, tools MAY use the value of the `actions` key (e.g. `io.cnab.dry-run`). User agents that display the display name SHOULD disambiguate if two or more actions have the same `display` value.
```
silvin-lubecki: nit: We use `title` in the `definitions` specs, which comes from JSONSchema, see https://github.com/cnabio/cnab-spec/blob/36d7b4817de6902142fc338549c8d93aa94b37a5/101-bundle-json.md#L411

technosophos:
I have updated this to use `title` instead of `display`. I agree that sounds much better.

schema/bundle.schema.json
```diff
@@ -187,6 +187,10 @@
             "description": "A description of the purpose of this action",
             "type": "string"
           },
+          "display": {
```
vdice: This should change to `title`, correct?

101-bundle-json.md
```diff
@@ -764,6 +767,7 @@ The above declares three actions: `io.cnab.status`, `io.cnab.migrate` and `io.cn
 
 Each action is accompanied by a description, which contains the following fields:
 
+- `title`: The human-readable name of the operation. If omitted, tools MAY use the value of the `actions` key (e.g. `io.cnab.dry-run`). User agents that display the title SHOULD disambiguate if two or more actions have the same `title` value.
```
vdice: Should we add `(OPTIONAL)` at the end of this entry, as is done elsewhere?  (Technically it looks like strict adherence isn't already in place as `modifies` below also doesn't have this...)

schema/bundle.schema.json
technosophos: Ugh... I missed a file in my commit. Sorry. Will fix that.

101-bundle-json.md
technosophos: Yes. Good catch.

Issue #215, opened 2019-07-05
CNAB Spec patterns are not RE2


astrieanna:
The [pattern for credential paths](https://github.com/deislabs/cnab-spec/blob/master/schema/bundle.schema.json#L19) and the [pattern for parameter paths](https://github.com/deislabs/cnab-spec/blob/master/schema/bundle.schema.json#L184) in `bundle.scheme.json` both use the `?!` lookahead feature that is allowed by jsonspec but is not available in Go's re2 regexes. Javascript and PCRE regexes support `?!`, but re2 has consistently responded to issues with no plans to do so.

Unfortunately, there is no obvious way to exclude a prefix that's more than one character long without using `?!`. In re2, the easiest thing would be to mark the pattern as a negative pattern -- "must not match" rather than the usual "must match". However, since javascript supports `?!`, it seems unlikely that jsonspec would want to add extra fields to support re2 implementation constraints.

It is valuable to have these checks in the json schema rather than just in markdown prose, but our golang implementation can't handle the patterns because they include `?!`.

cc @ryanmoran

technosophos:
Given that this is the spec repo... on the spec side we could merely drop the regex, or we can keep it and let the Go-based projects solve this problem on their own. I think I would be in favor of the second approach.

For Duffle/cnab-go, I guess the question is: Is there a PCRE implementation in Go? I guess we could use `cgo` to link it in... but that seems like a heavy-handed approach.

silvin-lubecki:
We have the same issue here on docker app. We can't use the json schema provided by the specs to validate the bundles, as AFAIK there's no ECMA-262 compatible regexp implementation in Go, and all the json schema implementations in Go use the regexp package provided by Go.
So even just parsing the schema is not possible... I'd be more in favor with removing it from the schema and add a proper validation in the cnab-go library.

technosophos:
If anyone is interested in replacing this, can you create a PR? If it can be done without breaking any existing CNAB modules, we can include it in CNAB Core 1.1.0

vdice:
Closed by https://github.com/cnabio/cnab-spec/pull/330

PR #330, opened 2020-02-14
ref(bundle.schema.json): remove pattern property from param/cred path

vdice:
Removes the `pattern` property from the definitions for parameters and credentials.

The negative lookup regex used proved problematic for validation (namely, incompatible with golang implementations, of which cnab's most-used - https://github.com/cnabio/cnab-go - is one).

This change would therefore mean implementations are responsible for enforcing spec expectations as regards these path strings, namely, `"Specified path MUST NOT be a subpath of /cnab/app/outputs."` as seen in https://github.com/cnabio/cnab-spec/blob/master/101-bundle-json.md.

Potential solution to https://github.com/cnabio/cnab-spec/issues/215

Corresponding cnab-go implementation: https://github.com/cnabio/cnab-go/pull/180

### DIFF
--- a/schema/bundle.schema.json
+++ b/schema/bundle.schema.json
@@ -16,7 +16,6 @@
         },
         "path": {
           "description": "The path inside of the invocation image where credentials will be mounted",
-          "pattern": "^(?!/cnab/app/outputs)/.+$",
           "type": "string"
         },
         "required": {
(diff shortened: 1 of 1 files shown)


vdice:
@technosophos I added this to the CNAB Core 1.1.0 milestone.  Assuming it gains consensus, it can be merged to master in the meantime, correct?  I think CNAB Core 1.0.1 is still itself pending (requires a vote), so I'm not sure what the 1.1.0 timeline would be.

technosophos:
@vdice My reading of our versioning is that this is not a change to the specification, properly speaking. So we can merge it into 1.0.1 if you would like.

PR #332, opened 2020-02-18
Establishing trust for multiple registries

### LINKED ISSUE #314: Document security implications of using multiple servers to host metadata repositories
This relates to CNAB Security 301: Metadata repositories as described in https://github.com/cnabio/cnab-spec/pull/281. See https://github.com/cnabio/cnab-spec/pull/281#discussion_r363237658.

A usecase I have in mind is where a user consumes bundles from multiple vendors who do not share a single, central server. I expect this to be the case if CNAB becomes popular and is adopted by multiple vendors who will not necessarily be able to agree on a central metadata repository.

In particular, it would be good to spell out how to establish trust in a new metadata repository - presumably mainly by reputation and guess-work? I believe there is a weakness of the current approach in the above usecase and it would be good to call that out so that there are no surprises down the line.

trishankatdatadog:
An attempt to fix #314


trishankatdatadog:
He probably does. I'll wait for him to say what he'd like to see.

trishankatdatadog:
> The example approaches given based on TAP 4 seem to require a willingness on the part of vendors to support metadata repositories other than their own or to agree to share a meta metadata repository. This then begs the same question: how does a vendor establish trust in a shared repository?

Hi Glyn, sorry, I don't fully understand. Does the following paragraph not answer the Q?

>  Trust for different metadata repositories with different roots of trust can be established out-of-band using an approach like [TAP 4](https://github.com/theupdateframework/taps/blob/master/tap4.md). For example, a bundle runtime could be shipped with known good copies of TUF `root` metadata for different repositories, or there could even be a meta metadata repository that distributes these `root` metadata. How to establish whether a `root` metadata file is trustworthy is out of the scope of this document.

Are you asking specifically how do we trust a `root` metadata file at all?

glyn:
> > The example approaches given based on TAP 4 seem to require a willingness on the part of vendors to support metadata repositories other than their own or to agree to share a meta metadata repository. This then begs the same question: how does a vendor establish trust in a shared repository?
> 
> Hi Glyn, sorry, I don't fully understand. Does the following paragraph not answer the Q?

I don't think so.

> 
> > Trust for different metadata repositories with different roots of trust can be established out-of-band using an approach like [TAP 4](https://github.com/theupdateframework/taps/blob/master/tap4.md). For example, a bundle runtime could be shipped with known good copies of TUF `root` metadata for different repositories, or there could even be a meta metadata repository that distributes these `root` metadata. How to establish whether a `root` metadata file is trustworthy is out of the scope of this document.
> 
> Are you asking specifically how do we trust a `root` metadata file at all?

Perhaps, if that is essentially the same thing as establishing trust in a metadata repository. Pardon my ignorance...

trishankatdatadog:
> Perhaps, if that is essentially the same thing as establishing trust in a metadata repository. Pardon my ignorance...

Not a problem. Shall we discuss this in the Security / Registry meeting next week, so that I am super clear about what you mean? Thanks!

trishankatdatadog:
I have added this action item to the meeting agenda...

glyn:
@trishankatdatadog Unfortunately my Wednesday evenings are currently spoken for, so I can't attend the meeting. Happy to chat some other time though. Maybe grab me on slack?

301-metadata-repositories.md
```diff
@@ -21,6 +21,7 @@ Note that:
...
 * Both TUF and in-toto are frameworks that can be configured in a wide variety of ways to achieve varying degrees of security. However, we discuss one [minimum viable product (MVP)](#minimum-viable-product-mvp), which has been optimized in particular for [Docker Content Trust / Notary 0.6.1](https://github.com/theupdateframework/notary/releases/tag/v0.6.1), in this document.
 * Although we discuss only the signing and verification of bundles, exactly the same principles apply to images.
+* Trust for different metadata repositories with different roots of trust can be established out-of-band using an approach like [TAP 4](https://github.com/theupdateframework/taps/blob/master/tap4.md). For example, a bundle runtime could be shipped with known good copies of TUF `root` metadata for different repositories, or there could even be a meta metadata repository that distributes these `root` metadata. How to establish whether a `root` metadata file is trustworthy is out of the scope of this document.
```
glyn: @trishankatdatadog here's a suggestion which would address my concerns:
```suggestion
* Trust for different metadata repositories with different roots of trust can be established out-of-band using an approach like [TAP 4](https://github.com/theupdateframework/taps/blob/master/tap4.md). For example, a bundle runtime could be shipped with known good copies of TUF `root` metadata for different repositories, or there could even be a meta metadata repository that distributes these `root` metadata.
* How to establish whether a `root` metadata file is trustworthy is outside the scope of this document. Simply assuming a `root` metadata file is trustworthy (e.g., by "clicking through") voids all guarantees (of end-to-end authenticity and integrity of bundles) provided by the corresponding metadata repository.
```

trishankatdatadog:
@glyn @jlegrone @radu-matei Are we good to go?

301-metadata-repositories.md
```diff
@@ -21,6 +21,8 @@ Note that:
...
 * Although we discuss only the signing and verification of bundles, exactly the same principles apply to images.
+* Trust for different metadata repositories with different roots of trust can be established out-of-band using an approach like [TAP 4](https://github.com/theupdateframework/taps/blob/master/tap4.md). For example, a bundle runtime could be shipped with known good copies of TUF `root` metadata for different repositories, or there could even be a meta metadata repository that distributes these `root` metadata.
+* How to establish whether a `root` metadata file is trustworthy is outside the scope of this document. Simply assuming a `root` metadata file is trustworthy (e.g., by "clicking through") voids all guarantees (of end-to-end authenticity and integrity of bundles) provided by the corresponding metadata repository.
```
radu-matei: I would note that this falls outside the scope of this document, but not not necessarily include the last sentence.

trishankatdatadog: @glyn Do you agree? In any case, please make a suggestion.

glyn: I think it's important to be up front about the consequences of failing to establish the necessary trust. I looked into moving the text to 300-CNAB-Security.md, but since it's specific to metadata repositories, it doesn't really fit there.

@radu-matei why do you think the last sentence should be omitted?

trishankatdatadog:
@glyn @radu-matei so should we merge, or?

technosophos:
I believe this is good to go now. @radu-matei if you agree, can you merge?

radu-matei:
Yeah, this only needs a rebase on top of `master` and we can merge.

trishankatdatadog:
@radu-matei @technosophos I have rebased. Thanks very much!

PR #310, opened 2020-01-09
Change 'name' to 'installation' on Claim

technosophos:
This changes the `name` field on a claim to be `installation`

BREAKING CHANGE: This will require updates to both the Rust and Go CNAB libraries

Depends upon #309 
Closes #303

PR #333, opened 2020-02-25
fix(claim.schema.json): s/name/installation in required section

vdice:
* Updates the `required` section of the `claim` schema to require `installation` instead of `name` (see https://github.com/cnabio/cnab-spec/pull/310 for genesis of this change)
* Updates example


vdice:
(GH appears to have spotty webhook transmission currently; hence pending CI)

PR #325, opened 2020-02-14
fix: change "underway" to "pending" in CNAB claims

technosophos:
This PR changes "underway" to "pending". Since this changes the behavior of cnab-go and libcnab-rs, we need to create issues in those repos if we decide to merge this.

See https://github.com/cnabio/cnab-spec/pull/309#discussion_r365532611 for discussion of what the name of this field should be.

Depends on #309

PR #334, opened 2020-02-25
change 'underway' to 'pending' in the claims spec

technosophos:
This PR changes "underway" to "pending". Since this changes the behavior of cnab-go and libcnab-rs, we need to create issues in those repos if we decide to merge this.

See #309 (comment) for discussion of what the name of this field should be.

Supersedes #325 


technosophos:
Turns out that no change is necessary in the Rust library, which currently allows any String value in the Status field. For Go, a PR is here: https://github.com/cnabio/cnab-go/pull/185

PR #339, opened 2020-02-28
Split claim into execution inputs and results

### LINKED ISSUE #326: Claims: It is unclear what happens to outputs on various operations
A question was raised as to what happens to outputs operation to operation: https://github.com/cnabio/cnab-spec/pull/309#discussion_r366716640

@jeremyrickard says:

> I think we should expand this normative text to describe what happens between actions.
> The text currently says:
>>   If this field is not present, it may be assumed that no outputs were generated as a result of the operation.

> What happens to this value between executions? If we address this bug in Porter: https://github.com/deislabs/porter/issues/836
> What is the correct answer? I think this spec doesn't provide enough prescriptive info about what the claim should provide.

For my part, I am not sure why outputs would persist from one to the next. That seems to me to be a misuse of a claim, which tracks only the state of the last operation. (It is not intended to be a general-purpose data storage mechanism.) To track older outputs, wouldn't you just store older claims instead of overwriting the claim?

There has also been some movement toward tracking outputs separately from claims, which could prove a more elegant solution.

### LINKED ISSUE #329: Claims: Move 'outputs' into 'results'
It has been proposed here https://github.com/cnabio/cnab-spec/pull/309#discussion_r365505351 that we move `outputs` into `results`.

@jlegrone says:

>I feel that action makes the most sense as a top-level field on the claim object, while outputs could potentially be moved into result. This has a practical benefit too; when a user invokes an operation against an external API, the server would be able to send back a stream of claim.result type events until execution is complete. 

Note that this will break cnab-go and libcnab-rust, so needs follow up issues if we decide to do it.

Note that this conflicts with #327

carolynvs:
Any fields that are modified based on the bundle execution: outputs, modified have moved to claim.results. The action, e.g. "install", has always been an input and has moved out of result to the top level claim.

Now claim (with the exception of its child result) does not change after creation. A tool could create a digest from it if needed.

Also results.outputs has been modified to only contain the output names instead of the entire output contents, which can be prohibitively large. The tool can request the outputs at the end of the operation to persist them separate from the claim. 

Closes #329 - Move output names into the results sub-document
Closes #238 - Move action to the top level claim document
Closes #326 - Store outputs outside of the claim

This is a breaking change to Porter at least, which relied on having the outputs in the claims to make bundle dependencies work. A follow-up change to the dependency spec will be necessary.


400-claims.md
```diff
@@ -126,16 +127,17 @@ The parameter data stored in a claim data is _the resolved key/value pairs_ that
 
 ### Outputs
 
-The output data saves the contents of each output file at the end of the action. The data is a map from output name as defined in the bundle to the content of the file located at the path defined for that output.
+The claim `result` provies a list of output names generated by the operation. A tool may choose to request the contents of any outputs to persist them but is not required to do so.
 
-Below, you can see an example of a claim for a bundle that included a single output file, `clientCert`. The value in the claim's output data is the contents of `/cnab/app/outputs/clientCert` at the end of the `install` action.
+The output name, as defined in the bundle, is used to request the content of the file located at the path defined for that output.
+
+Below, you can see an example of a claim result that included an output file, `clientCert`.
```
vdice:
```suggestion
Below, you can see an example of a claim result that includes an entry for the output `clientCert`.
```

schema/claim.schema.json
```diff
@@ -44,17 +52,13 @@
...
-      "type": "string"
-    },
     "installation": {
       "description": "the name of the installation",
       "type": "string"
     },
-    "outputs": {
-      "description": "key/value pairs of output name to output value.",
-      "type": "object"
+    "modified": {
```
vdice: This `modified` property should move up into the `result` definition, right?

400-claims.md
```diff
@@ -84,22 +84,23 @@ Source: [400.01-claim.json](examples/400.01-claim.json)
...
 - `result` (REQUIRED): The outcome of the bundle's last action (e.g. if action is install, this indicates the outcome of the installation.). It is an object with the following fields:
   - `action` (REQUIRED): The name of the action that last caused this claim to be updated. This may be any of the built-in actions (`install`, `upgrade`, `uninstall`) as well as any custom actions as defined in the bundle descriptor. The special name `unknown` MAY be used in the case where the CNAB Runtime cannot determine the action name of a claim.
   - `message` (OPTIONAL): A human-readable string that communicates the outcome. Error messages MAY be included in `failure` conditions.
+  - `outputs` (OPTIONAL): List of output names generated by the operation. These are stored so that the user can access them after the operation completes. Some implementations MAY choose not to store these for security or portability reasons. See the [Outputs](#outputs) section for details. If this field is not present, it may be assumed that no outputs were generated as a result of the operation.
```
vdice: It appears we don't have an entry for the `modified` property here under `result`... it may have been lost when removing from the main claim properties section.

400-claims.md
```diff
@@ -84,22 +84,23 @@ Source: [400.01-claim.json](examples/400.01-claim.json)
...
   - `status` (REQUIRED): Indicates the status of the last phase transition. Valid statuses are:
-    - `failure`: failed before completion
-    - `pending`: in progress. This should only be used if the invocation container MUST exit before it can determine whether all operations are complete. Note that `pending` is a _long term status_ that indicates that the installation's final state cannot be determined by the system. For this reason, it should be avoided. When used, `pending` should be considered a temporary status, and the runtime SHOULD work to resolve this to either `failure` or `success`.
-    - `unknown`: the state is unknown. This is an error condition.
-    - `success`: completed successfully
+    - `failure`: Failed before completion.
+    - `pending`: Execution has been requested and has not begun.
+    - `running`: Execution is in progress and has not yet completed.
```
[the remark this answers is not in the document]
carolynvs: Oh shoot, sorry. This branch originally had all the changes together that came out of our claim WG, and then this snuck in. Thanks for catching that!

schema/claim.schema.json
carolynvs: Thanks, yes my git-fu went sideways this morning.

400-claims.md
```diff
@@ -84,17 +84,17 @@ Source: [400.01-claim.json](examples/400.01-claim.json)
...
 - `result` (REQUIRED): The outcome of the bundle's last action (e.g. if action is install, this indicates the outcome of the installation.). It is an object with the following fields:
-  - `action` (REQUIRED): The name of the action that last caused this claim to be updated. This may be any of the built-in actions (`install`, `upgrade`, `uninstall`) as well as any custom actions as defined in the bundle descriptor. The special name `unknown` MAY be used in the case where the CNAB Runtime cannot determine the action name of a claim.
   - `message` (OPTIONAL): A human-readable string that communicates the outcome. Error messages MAY be included in `failure` conditions.
+  - `modified` (OPTIONAL): A timestamp indicating the last time this result was modified. Executing the installation action MUST set this to a timestamp that matches the claim's `created`. Executing any modifying action (`upgrade`, `uninstall`, or a custom action with the property `"modifies": true`) MUST set this field to the time of the operation.
```
jlegrone: I think it would be useful to track start and end time of operations, but `modified` doesn't quite provide enough information to handle this. Could we add something like `startTime` and `endTime`?

carolynvs:
When @vdice and I were discussing the retry case, we were having a hard time understanding whether or not a retry would be able to reuse a claim as-is (i.e. not change the revision):

> revision (REQUIRED): An ULID that MUST change each time the claim is modified. It MUST NOT change when a non-modifying operation is performed on an installation.

from [Claims Spec](https://github.com/cnabio/cnab-spec/blob/master/400-claims.md)

Now that we are working to make all of the fields on the claim (other than result) immutable, this text is not longer clear.

400-claims.md
carolynvs: Oh yeah I had that in my notes and forgot to add it. 👍

jlegrone:
> When @vdice and I were discussing the retry case, we were having a hard time understanding whether or not a retry would be able to reuse a claim as-is (i.e. not change the revision)

I think there are several reasons why a claim could be reused, and that it should be up to the runtime to determine whether this is the right approach. For instance, the Kubernetes driver could return an error result when there are no compute resources available in the cluster to run an invocation image, but a higher-order driver could then pass off the same claim to a secondary driver instance running in a different cluster.

Another use case would be to allow a client to retry stateless or modifies=false actions without generating a new claim. Stateless action results could even be cached by the runtime in some cases, though the spec doesn't require invocation image outputs to be deterministic.

400-claims.md
technosophos: Just to be clear: `modified` is the time the claim was last modified, not the time of an operation. I think `started` and `stopped` below are relevant to the runtime's execution time, right?

technosophos:
The purpose of a claim is to _record an event/action that mutates the resources of a cluster as a result of a CNAB operation_.

Any operation that does not mutate the installed sources must not generate a new claim (e.g. if an action has modifies=false, it must not generate a claim, and this is probably true of all stateless commands as well). Claims are to record any case where the managed artifacts were modified by a CNAB tool. (If I need to clarify this more, I will. I can see how this might not be totally clear from 400 alone)

Ultimately, it is up to the runtime to determine whether it will "reuse a claim", but things will get awfully confusing if some systems decide not to record new claims when they change installations.

jlegrone:
> Any operation that does not mutate the installed sources must not generate a new claim (e.g. if an action has modifies=false, it must not generate a claim, and this is probably true of all stateless commands as well). Claims are to record any case where the managed artifacts were modified by a CNAB tool. (If I need to clarify this more, I will. I can see how this might not be totally clear from 400 alone)

I'm not sure this works given the current model of one claim embedding one result. What should a runtime do if a user runs a modifies=false action that generates an output? Modify the last claim? Do we want the spec to lay out how runtimes should maintain an audit log of all actions that were executed against an installation, including stateless and modifies=false?

technosophos:
I am not convinced that outputs/results really do belong in claims. I would rather seem them as separate. The big question for me is what do we want the relationship to be between claims and results? Can we just have a claim point to an output (or vice-versa)?

My database brain does not like the idea of just nesting one inside the other, because regardless of which way we structure it, we will be saying one of them "belongs" to the other, which isn't necessarily true. In my mind, while the claim is intended to capture the responsibility for a change to resources, a "result" is merely the stored output of any operation (regardless of whether it changed anything).

One could imagine a system that stored claims, but not results.

One can imagine cases where one is only really concerned with results from non-mutating actions, but not really with claims at all.

All of that said, I'm certainly open to being convinced otherwise.

400-claims.md
carolynvs: Modified is when the claim result was last modified. We moved it to result since nothing on the parent claim changes after the claim is created.

Start and Stop are for the execution time, correct.

carolynvs:
Perfect, yeah let's find a way for result to live on its own in a separate PR then. I'm not a codeowner, so once y'all are happy with this one, would someone else merge please?

PR #340, opened 2020-02-28
Clarify pending status, introduce running

### LINKED ISSUE #312: Claims: Store a claim when an action has started
From @carolynvs  #251:

Providing a way to indicate on a claim that an action has begun. Consider the prototype CNAB Dashboard I showed a while back. If we updated the claim when an action has started, you could know which bundles are being currently executed, and which action. Right now claims are stored on the local file system so it may not make sense, but if you think about in the future when teams may have claims persisted in a common location, and are reporting off of that claim store... That's where this would be useful.
https://github.com/cnabio/cnab-spec/issues/251#issuecomment-529001842

carolynvs:
The current definition for the pending (previously underway) status is difficult to understand when it is acceptable to use and what it really means about the state of the installation. It's described as both a long term state and temporary. It is also described as "to be avoided".

We would like to redefine the state to make it more clear on what the state means, and to make it acceptable for tools to use. Pending now means that the operation has been requested (most likely queued) but has not begun executing. Tools that are immediately executing commands as
they are issued would most likely choose to not use this state. However solutions that accept bundle operations from many users at once, for example a cnab operator, may choose to use this state.

This also introduces a new state: running. Tools may use this state to indicate that a bundle is currently executing an operation. This can be used by tools to provide operational insight into the activity of the system. For example, when an upgrade takes 30 minutes and people don't
know that it's already running, it could easily be initiated more than once by multiple people on the same installation. Allowing tools to make the current status of the installation visible, helps avoid this situation.

Closes #312

### DIFF
--- a/400-claims.md
+++ b/400-claims.md
@@ -99,8 +99,8 @@ The fields above are defined as follows:
   - `started` (OPTIONAL): A timestamp indicating when execution of the operation started.
   - `status` (REQUIRED): Indicates the status of the last phase transition. Valid statuses are:
     - `failure`: Failed before completion.
-    - `pending`: Execution has been requested and has not begun.
-    - `running`: Execution is in progress and has not completed.
+    - `pending`: Execution has been requested and has not begun. This should be considered a temporary status, and the runtime SHOULD work to resolve this to either `failure` or `success`.
+    - `running`: Execution is in progress and has not completed.  This should be considered a temporary status, and the runtime SHOULD work to resolve this to either `failure` or `success`.
     - `unknown`: The state is unknown. This is an error condition.
     - `success`: Completed successfully.
   - `stopped` (OPTIONAL): A timestamp indicating when the execution of the operation stopped.
--- a/400-claims.md
+++ b/400-claims.md
@@ -98,10 +98,11 @@ The fields above are defined as follows:
   - `revision` (REQUIRED): An [ULID](https://github.com/ulid/spec) that MUST change each time the claim is modified. It MUST NOT change when a [non-modifying operation](https://github.com/cnabio/cnab-spec/blob/master/101-bundle-json.md#custom-actions) is performed on an installation.
   - `started` (OPTIONAL): A timestamp indicating when execution of the operation started.
   - `status` (REQUIRED): Indicates the status of the last phase transition. Valid statuses are:
-    - `failure`: failed before completion
-    - `pending`: in progress. This should only be used if the invocation container MUST exit before it can determine whether all operations are complete. Note that `pending` is a _long term status_ that indicates that the installation's final state cannot be determined by the system. For this reason, it should be avoided. When used, `pending` should be considered a temporary status, and the runtime SHOULD work to resolve this to either `failure` or `success`.
-    - `unknown`: the state is unknown. This is an error condition.
-    - `success`: completed successfully
+    - `failure`: Failed before completion.
+    - `pending`: Execution has been requested and has not begun.
+    - `running`: Execution is in progress and has not completed.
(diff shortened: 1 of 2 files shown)


carolynvs:
> Would a tool be expected to generate a new claim and result with this "conclusive" status? Or just modify the result on/associated with an existing claim?

The expectation is result is a modifiable sub-document, it contains fields that are outputs from the execution of the bundle (per the other PR), and they can be set without requiring a new claim to be created.

vdice:
> and they can be set without requiring a new claim to be created.

This makes sense.  In a follow-up, we might consider adding specification that status updates from transient to completed statuses either should or must update the result object and either should not or must not update the associated claim or claim revision...

400-claims.md
```diff
@@ -96,10 +96,11 @@ The fields above are defined as follows:
...
   - `message` (OPTIONAL): A human-readable string that communicates the outcome. Error messages MAY be included in `failure` conditions.
   - `status` (REQUIRED): Indicates the status of the last phase transition. Valid statuses are:
-    - `failure`: failed before completion
-    - `pending`: in progress. This should only be used if the invocation container MUST exit before it can determine whether all operations are complete. Note that `pending` is a _long term status_ that indicates that the installation's final state cannot be determined by the system. For this reason, it should be avoided. When used, `pending` should be considered a temporary status, and the runtime SHOULD work to resolve this to either `failure` or `success`.
-    - `unknown`: the state is unknown. This is an error condition.
-    - `success`: completed successfully
+    - `failure`: Failed before completion.
+    - `pending`: Execution has been requested and has not begun.
```
silvin-lubecki: >  When used, `pending` should be considered a temporary status, and the runtime SHOULD work to resolve this to either `failure` or `success`.

I like this new `running` state, but I think we should keep the part enforcing that this status (and also `pending`) is only temporary and that the tools should resolve it .

technosophos:
Follow-up issues should be filed in cnab-go and libcnab-rust

400-claims.md
```diff
@@ -96,10 +96,11 @@ The fields above are defined as follows:
...
   - `message` (OPTIONAL): A human-readable string that communicates the outcome. Error messages MAY be included in `failure` conditions.
   - `status` (REQUIRED): Indicates the status of the last phase transition. Valid statuses are:
-    - `failure`: failed before completion
-    - `pending`: in progress. This should only be used if the invocation container MUST exit before it can determine whether all operations are complete. Note that `pending` is a _long term status_ that indicates that the installation's final state cannot be determined by the system. For this reason, it should be avoided. When used, `pending` should be considered a temporary status, and the runtime SHOULD work to resolve this to either `failure` or `success`.
-    - `unknown`: the state is unknown. This is an error condition.
-    - `success`: completed successfully
+    - `failure`: Failed before completion.
```
jlegrone: It might be more consistent to use an adjective for each status

```suggestion
    - `failed`: Failed before completion.
```

400-claims.md
```diff
@@ -96,10 +96,11 @@ The fields above are defined as follows:
...
+    - `pending`: Execution has been requested and has not begun. This should be considered a temporary status, and the runtime SHOULD work to resolve this to either `failure` or `success`.
+    - `running`: Execution is in progress and has not completed.  This should be considered a temporary status, and the runtime SHOULD work to resolve this to either `failure` or `success`.
+    - `unknown`: The state is unknown. This is an error condition.
+    - `success`: Completed successfully.
```
jlegrone:
```suggestion
    - `succeeded`: Completed successfully.
```

carolynvs:
I will open a follow-on PR to standardize our verbs and add cancelled.

PR #343, opened 2020-03-04
painstakingly explained versioning

### LINKED ISSUE #342: Define version compatibility behavior
Now that we have a new version of the core spec, 1.0.1, what behavior can people rely upon when the version that the bundle was built against doesn't match what the tool they are using to run it supports?

Examples:

* **Forwards Compatibility** They are still using an old duffle binary that was built for 1.0.0 and are running a bundle built for 1.0.1
* **Backwards Compatibility** They are using a bundle built for 1.0.0 and the latest duffle which supports 1.0.1

The behavior should be defined in https://github.com/cnabio/cnab-spec/blob/master/901-process.md#versioning

technosophos:
Much of the CNAB versioning regime was "tribal knowledge" or explained tersely in the governance documents. This codifies the versioning scheme.

Closes #342 

### DIFF
--- a/901-process.md
+++ b/901-process.md
@@ -34,9 +34,12 @@ A small number of stability markers are allowed, as determined by [the foundatio
 
 - `PD`: Pre-draft indicates that the version of the resource is an unstable testing version.
 - `DRAFT`: Draft indicates that the version is an unstable in-development version
-- `GA`: Working Group Approval indicates that this version has been accepted by the maintainers, but not accepted by the Executive Directors. In practice, this marker MUST only be used internally for testing, and never for production. 
+- `GA`: Working Group Approval indicates that this version has been accepted by the maintainers, but not accepted by the Executive Directors. In practice, this marker MUST only be used internally for testing, and never for production.
+- `AD`: Final approval by the Executive Directors.
 
-The tags `GA` and `AD` should never be used in a SemVer stability marker.
+The tags `GA` and `AD` should never be used in a SemVer stability marker. `GA` is not a milestone that warrants publication, and `AD` is synonymous with the final release.
+
+The development process defines `Publication` and `Submission to Other Standards Bodies` as process steps. However, neither of these reflects a change in the specification. These are people processes.
 
 The stability markers `ALPHA`, `BETA`, and so on are _disallowed_ under this specification, and MUST NOT be used to express CNAB versions.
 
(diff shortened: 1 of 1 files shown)


901-process.md
```diff
@@ -15,6 +15,38 @@ For example, one of the specifications in this repository is the CNAB Core speci
...
+- Patch releases (`Z`): Patch release contain fixes to the text of the specification. Patch releases MUST NOT change the behavior of the specification (except in cases where the specification was too vague and the patch clarifies).
+    - Patch releases MUST be both forward and backward compatible to the minor version number
+    - Patch releases MUST NOT make the schema harder to validate against (though they may relax the schema).
+    - Our intention is that CNAB Runtimes SHOULD NOT have to specify behavior at the patch level, as all patch levels are compatible.
+
+Stability markers provide a way to indicate that a bundle or runtime is experimenting with features or fixes. _If an object is tagged with a stability marker, it MUST be treated as incompatible with any other version number. E.g. `1.0.0-GA` MUST be considered incompatible with `1.0.0`. Production artifacts SHOULD NOT use stability markers.
+
+A small number of stability markers are allowed, as determined by [the foundation's governance documents](https://github.com/cnabio/community/blob/master/governance.md):
```
radu-matei: Interestingly enough, the security spec uses `-WD` (see [table of contents in the readme of the spec repo](https://github.com/cnabio/cnab-spec#table-of-contents)) - which does not appear to be part of the allowed stability markers.

Should we update that `-DRAFT`?
(Also, I recall we used `-WD` extensively in the past as well for other specs).

901-process.md
```diff
@@ -15,6 +15,38 @@ For example, one of the specifications in this repository is the CNAB Core speci
...
+
+- `PD`: Pre-draft indicates that the version of the resource is an unstable testing version.
+- `DRAFT`: Draft indicates that the version is an unstable in-development version
+- `GA`: Working Group Approval indicates that this version has been accepted by the maintainers, but not accepted by the Executive Directors. In practice, this marker MUST only be used internally for testing, and never for production. 
+
```
radu-matei: This is a bit unclear - the linked governance document contains more phases than the three stability markers described here - are all phases valid markers in versions?

technosophos: No -- only the ones here. I will clarify that.

technosophos: Fixed.

PR #344, opened 2020-03-04
Standardize claim status verbs

carolynvs:
* Use past tense for final states.
* Use present participle, *-ing, for temporary states.
* Unknown is magic, and not a verb. It defies all standards.

This is a follow-up to https://github.com/cnabio/cnab-spec/pull/340/files#r387812107


carolynvs:
😂 All my knowns are unknown and that's the way I like it

PR #347, opened 2020-03-06
Revert moving claims.revision

carolynvs:
In #339 I accidentally changed the indent level of revision and moved it under `result` in the field defitions. It is defined in the top level claim document, and this fixes the indentation back to where it was before.


silvin-lubecki:
LGTM

PR #349, opened 2020-03-09
Split claim and claim result into separate documents

### LINKED ISSUE #328: Claims: Move 'action' field to top level
It was proposed here https://github.com/cnabio/cnab-spec/pull/309#discussion_r365505351 that the `action` field be moved.

@jlegrone says:

> I feel that action makes the most sense as a top-level field on the claim object

Note that this will break cnab-go and libcnab-rust, so if we decide to do this, we should create issues there

carolynvs:
This is follow-up to #339 
* Split the claim document into two documents: Claim and Claim Result. A claim may have multiple results if it is attempted multiple times (for example it is cancelled and retried).
* I noticed that result `message` was marked as required in the schema but optional in the spec document, and was optional in cnab-go. I added a commit that changes the result message to optional. If it was intended to be required, I can flip that commit around to update the spec text instead.
* When claim and result were on the same document, we briefly considered
splitting `created` and `modified` between the top level claim and the result.
Undo that change and have a created/modified pair for both claim and
result. Tools may choose to treat claim as immutable, but that isn't defined in
the spec.
* Mark both created/modified as required, matching the schema and how they
are used in cnab-go.
* Allow revision to be updated for non-modifying actions.

Closes #328


schema/claim-result.schema.json
```diff
@@ -0,0 +1,82 @@
+{
+  "$id": "https://cnab.io/v1/claim-result.schema.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "definitions": {
+    "claimRef": {
```
vdice: Should this `claimRef` have any required fields (`installation` or `revision` or both?) or acceptable to be empty?  (`claimRef` itself is designated as required on this claim-result)

schema/claim.schema.json
```diff
@@ -66,6 +19,10 @@
       "description": "The date created, as an ISO-8601 Extended Format date string, as specified in the ECMAScript standard",
       "type": "string"
     },
+    "modified": {
```
vdice: This is one area that I'm still grappling with... now that we separate the claim-result from the claim, with intentions that all updates/modifications are in the former, what scenarios will this claim object be modified (of, have a differing timestamp between `created` and `modified`)?

400-claims.md
```diff
@@ -128,64 +128,105 @@ The parameter data stored in a claim data is _the resolved key/value pairs_ that
...
+
+- `claimRef` (REQUIRED): Reference to the claim that generated this result.
+  - `installation` (REQUIRED): The claim installation name.
+  - `revision` (REQUIRED): The claim revision.
+- `custom` (OPTIONAL): A section for custom extension data applicable to a given runtime.
+- `id` (REQUIRED): An [ULID](https://github.com/ulid/spec).
+- `created` (REQUIRED): A timestamp indicating when this result was first created. This MUST not be changed after initial creation.
+- `modified` (REQUIRED): A timestamp indicating the last time this result was modified. Creating the result MUST set this to a timestamp that matches the result's `created`.
+- `message` (OPTIONAL): A human-readable string that communicates the outcome. Error messages MAY be included in `failed` conditions.
+- `outputs` (OPTIONAL): Map of output names generated by the operation. These are stored so that the user can access them after the operation completes. Some implementations MAY choose not to store these for security or portability reasons. See the [Outputs](#outputs) section for details. If this field is not present, it may be assumed that no outputs were generated as a result of the operation.
```
vdice: Is it worth mentioning that output values MAY be included as well, since the object itself is still a key-value map?  Or, is there agreement that actual output values should not be included (values always empty)?

schema/claim.schema.json
vdice: Ah, as noted in the PR description: `Tools may choose to treat claim as immutable, but that isn't defined in the spec.`  This helps.  So, a claim may still be mutable and hence `modified` might then be utilized (different from `created`).

400-claims.md
carolynvs: Oh, yeah let me clarify that output values are NOT in the claim result. The struct is reserved for future use (coming in another PR) containing metadata about the output. We want to put stuff in there like the content digest for example. But the full output value wouldn't be in there anymore.

schema/claim.schema.json
carolynvs: I believe that this is how those fields are used:

* claim.created = The timestamp of when the claim for the entire bundle _installation_ was first created. This never ever changes, even when new revisions are created.
* claim.modified = The timestamp for when the bundle _installation_ was last upgraded, or invoked with a custom action that modifies resources (or uninstalled). This is what is bumped when we make new revisions.

If you think about what `porter list` displays today, we are able to use created and modified to tell when the bundle was installed and last executed using those two fields. That won't change.

What I meant by that comment was that there are two ways to think about the claim document:

1. The way that duffle treats it, where we constantly overwrite the document and lose the data from each revision
2. The way that we have been encouraged to treat the document going forward (to solve the outputs being lost problem), as snapshots of data (installation + revision) that should be stored separately.

My comment was referring to the second way of looking at claims, where the claim document after this change could treat the claim document as immutable because all of the data captured when an operation is executed doesn't change afterwards because everything unknown at the time of creation was moved into result. Further modifications to the bundle installation, like doing another operation such as upgrade, would result in a new document being created with a bumped revision of the claim.

I didn't put that comment about "immutable" in the spec because this mental model  (overwriting vs separate documents) isn't advocated for in the spec. It says you can do it either way (though we've found that overwriting really doesn't work well). I just mentioned it in the PR comment because we have been talking about the second model quite a bit with this change (moving result out of claim). Sorry if I caused extra confusion!

vdice: Thank you for the detailed reasoning -- definitely helps me understand.

400-claims.md
```diff
@@ -86,24 +98,12 @@ The fields above are defined as follows:
 
 - `action` (REQUIRED): The name of the action that last caused this claim to be updated. This may be any of the built-in actions (`install`, `upgrade`, `uninstall`) as well as any custom actions as defined in the bundle descriptor. The special name `unknown` MAY be used in the case where the CNAB Runtime cannot determine the action name of a claim.
 - `bundle` (REQUIRED): The bundle, as defined in [the Bundle Definition](101-bundle-json.md).
-- `created` (OPTIONAL): A timestamp indicating when this release claim was first created. This MUST not be changed after initial creation.
+- `created` (REQUIRED): A timestamp indicating when this claim was first created. This MUST not be changed after initial creation and does not change with new revisions.
+- `modified` (REQUIRED): A timestamp indicating the last time this claim was modified. Executing the installation action MUST set this to a timestamp that matches the claim's `created`. Executing any modifying action (`upgrade`, `uninstall`, or a custom action with the property `"modifies": true`) MUST set this field to the time of the operation.
```
jlegrone: This seems to be the last mutable field on claims, and it seems like the `started` and `stopped` fields on result objects seem like they encode the same information. Could we remove `modified` and specify that claims are immutable instead?

400-claims.md
```diff
@@ -128,64 +128,105 @@ The parameter data stored in a claim data is _the resolved key/value pairs_ that
...
+### Claim Results
+
+The result of executing an operation is defined separately in a claim result document. A claim may have multiple results. For example if the operation was cancelled and then retried the claim will have two results: one representing the cancelled operation and another the retried operation.
 
-The claim `result` provides a list of output names generated by the operation. A tool may choose to request the contents of any outputs to persist them but is not required to do so.
+```json
+{
+  "claimRef": {
+    "installation": "technosophos.helloworld",
+    "revision": "01CP6XM0KVB9V1BQDZ9NK8VP29"
```
jlegrone: I don't think this is enough data to ensure that a runtime can resolve a single claim. For instance if two stateless actions are run with different parameters, this would result in two distinct claims but `installation` and `revision` would be the same between each of them.

I'm not sure what the best solution is here; the only ways to ensure correctness as far as I can tell are to embed the entire claim object, or to make claims immutable and reference by content digest. Content digests would also come with the benefit of keeping potentially sensitive data out of results and being small relative to full claim objects.

400-claims.md
carolynvs: I think that modified is a field that could be used in different ways depending on how the tool stores claim documents. (See [my comment above](https://github.com/cnabio/cnab-spec/pull/349#discussion_r390486137) for more context on the two ways to store documents. I kept a modified field because of questions I received about removing modified entirely from the claim document, there wasn't clear consensus about making it immutable.

I believe this leaves enough room each implementer to decide if it wants to treat the claim document as immutable or not (overwriting the claim document on each revision or persisting each revision as a separate document)? Otherwise, let's talk about this at the next meeting and have people talk it out. 😀

400-claims.md
carolynvs: Ah shoot, we completely missed that there isn't anything to reference for `modifies: false` when talking this over last week... Good catch.

400-claims.md
```diff
@@ -128,64 +128,105 @@ The parameter data stored in a claim data is _the resolved key/value pairs_ that
...
+  - `installation` (REQUIRED): The claim installation name.
+  - `revision` (REQUIRED): The claim revision.
+- `custom` (OPTIONAL): A section for custom extension data applicable to a given runtime.
+- `id` (REQUIRED): An [ULID](https://github.com/ulid/spec).
+- `created` (REQUIRED): A timestamp indicating when this result was first created. This MUST not be changed after initial creation.
+- `modified` (REQUIRED): A timestamp indicating the last time this result was modified. Creating the result MUST set this to a timestamp that matches the result's `created`.
+- `message` (OPTIONAL): A human-readable string that communicates the outcome. Error messages MAY be included in `failed` conditions.
+- `outputs` (OPTIONAL): Outputs generated by the operation. It is a map from the output names to metadata about the output. The  output value is not stored in the claim result. See the [Outputs](#outputs) section for details. If this field is not present, it may be assumed that no outputs were generated as a result of the operation.
+- `started` (OPTIONAL): A timestamp indicating when execution of the operation started.
```
jlegrone: Should we allow results with final statuses, like `succeeded` or `failed` to set this field? Would it be valid for other "still running" statuses to have the `stopped` field set?

400-claims.md
```diff
@@ -50,33 +50,45 @@ The CNAB claim is defined as a JSON document. The specification currently does n
...
+        "path": "/cnab/app/outputs/port"
+      }
+    },
     "parameters": {},
     "schemaVersion": "v1.0.0",
-    "version": "0.1.0"
+    "version": "0.1.2"
   },
-  "bundleReference": "hub.example.com/my/bundle@sha256:eeeeeeeee...",
+  "bundleReference": "hub.example.com/technosophos/hellohelm@sha256:eec03d9da1bad36b3e3a4526cc774153f7024a94f25df8d2dc3ca5602fc5273d",
```
technosophos: should probably remove my username from here, too. I thought I got all of those. Sorry.

400-claims.md
```diff
@@ -50,33 +50,45 @@ The CNAB claim is defined as a JSON document. The specification currently does n
...
   "created": "2018-08-30T20:39:55.549002887-06:00",
+  "modified": "2018-08-30T20:39:55.549002887-06:00",
   "custom": {},
-  "installation": "technosophos.hellohelm",
-  "modified": "2018-08-30T20:39:59.611068556-06:00",
-  "outputs": {},
+  "installation": "technosophos.helloworld",
```
technosophos: Ugh... and here. I need to stop copy/pasting.

technosophos:
Sorry about all the `technosophos` references. I did a purge on all of those, but must have accidentally re-introduced them on a subsequent merge. I will fix those in a follow-up PR. Don't worry about it.

400-claims.md
carolynvs: Sorry I missed this comment! 😊 Are you asking about `stopped` or `started`? I had intended for result to be mutable, so as the operation proceeded the tool would usually look like this:

1. Create the result and set created, modified to the same time. If it uses queuing, set status to `pending`. 
2. When execution starts set `started` and the status to `running`. At this point stopped is null.
3. When execution completes set `stopped` and status to `failed` or `succeeded`.

Based on how the tools works, some of these may occur in a single update to the result or in separate steps, but anytime it modifies the result, `modified` is bumped.

Does that help clarify? Does any of the wording in the PR need to change to reflect what I just said in the example above?

carolynvs:
I have pushed a new commit that changes `revision` so that implementations MAY update it for non-modifying actions. This should address https://github.com/cnabio/cnab-spec/pull/349#discussion_r391054936.

From the commit message:

---

To support results referencing back to claims for non-modifying actions
using only the installation name + revision, allow an implementation to
assign a new revision for non-modifying actions.

Implementations that intend to persist results for every action will use
this to assign a new revision to the claim for every action (modifying
or otherwise) to guarantee that that the reference is unique.
Implementations that chose to only persist the last modifying action
or only persist the revisions of modifying actions, can continue
assigning the revision as they do today.


This seems like a smaller change while allowing the implementations room to implement various aspects as they see fit (immutable, indexing by content digest, overwriting the claim file, keeping just the revisions of modifying actions like update but not stuff like logs, etc).

carolynvs:
Rebased to pick up the latest changes to the CI checks.

400-claims.md
technosophos: As I read the spec in its current version, this field can be treated as immutable.

- Install sets both to the same
- Upgrade clones the install time, and sets modified to the current time
- A non-destructive op clones both.

400-claims.md
```diff
@@ -104,7 +106,7 @@ The fields above are defined as follows:
...
 - `parameters` (OPTIONAL): Key/value pairs that were passed in during the operation. These are stored so that the operation can be re-run. Some implementations MAY choose not to store these for security or portability reasons. However, there are some caveats. See the [Parameters](#parameters) section below for details.
-- `revision` (REQUIRED): An [ULID](https://github.com/ulid/spec) that MUST change each time the claim is modified. It MUST NOT change when a [non-modifying operation](https://github.com/cnabio/cnab-spec/blob/master/101-bundle-json.md#custom-actions) is performed on an installation.
+- `revision` (REQUIRED): An [ULID](https://github.com/ulid/spec) that MUST change each time the claim is modified. It MAY change when a [non-modifying operation](https://github.com/cnabio/cnab-spec/blob/master/101-bundle-json.md#custom-actions) is performed on an installation.
```
jlegrone, line 109: Regarding the change from "MUST NOT" to "MAY":

We are using the current wording of the spec to enforce that 

1. No concurrent executions of `modifies: true` actions are run for the same installation (linearizability for modifying actions)
2. Any `modifies: true` action invocations are queued until all running `stateless: false` actions at the current revision are completed or canceled
3. Any number of `stateless: false, modifies: false` actions are allowed to run concurrently as long as there are no `modifies: true` actions in progress
4. Any claims for `stateless: false` or `modifies: true` actions referencing revisions older than the latest are invalidated and cannot be used to start new operations
5. Claims for `stateless: true, modifies: false` actions are always valid and may be executed concurrently with any other action(s)

Another thought is that in workflows that execute multiple `modifies: false` actions against the same installation, the revision may be used as a key to detect conflicts in a "transaction" and abort or retry.

We should be careful to avoid the implementation details of any particular runtime from leaking unnecessarily into the spec. Certainly points 1-2 can be handled with the proposed change, but I'm worried that in the future enforcing these additional invariants may be problematic for runtimes that have committed to always updating revisions.

400-claims.md
```diff
@@ -48,6 +48,8 @@ While implementors are not REQUIRED to implement claims, this is the standard fo
 
 The CNAB claim is defined as a JSON document. The specification currently does not require that claims be formatted as Canonical JSON.
 
+The latest revision of a claim MUST be retained. Previous revisions of a claim MAY be retained to provide a history of actions performed on an installation.
```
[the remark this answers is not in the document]
carolynvs: Why the restriction to only keeping the last "successful" installation revision? That is a deviation from how we currently retain the result of the last action in the current specification, ignoring this PR. 

Otherwise if I attempt an upgrade, and it fails, does that mean that we can just pretend it didn't happen and the last revision kept is still the successful installation? :-)

Otherwise I'm good with the suggested change, it's just that wording change (introducing allowing only retaining the last successful revision, as opposed to the revision) that I'm unsure about.

technosophos:
I believe this just needs a force-push sign-off from @carolynvs and we are ready to merge this, right? (Or are there still unaddressed issues?)

400-claims.md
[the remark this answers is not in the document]
jlegrone: I was thinking of a rollback use case; if a runtime must always keep track of the latest successful revision, then you'll always have the information you need in order to attempt an upgrade to that last known "good" state using the same bundle version + parameters.

In practice I think runtimes will choose to keep history for the last _n_ revisions anyway so I'm fine with the wording as is.

400-claims.md
jlegrone: Sorry for the late reply, just noticed this thread had more activity.

I'm a tiny bit confused now. Are we saying that `claim.created` represents the time of creation for the installation itself? I was thinking this was the time the claim was created. My working model is that the time of creation for the installation would be derived from the `claim.created` field of the claim with the earliest known revision.

Claim results _also_ have `created` and `modified` fields. This makes me question having `modified` on claims as well, because there is no standard way to answer questions like "when was this action completed". A runtime may choose do one of

- treat claims and results as immutable, returning the last known `result.created` timestamp associated with the claim
- treat claims as immutable but results as mutable, returning the `result.modified` timestamp
- treat claims as mutable, returning the `claim.modified`

I think more clarity on this from the spec would help improve the chances that implementers are able to build on the same core libraries in the future.

400-claims.md
jlegrone: That makes sense. I'm still wondering if we could still satisfy everyone's use cases without created, updated, started, and stopped timestamps all on the result object.

I would prefer to have a single `observed` timestamp and treat results as a list of events associated with a claim, but even if we don't require implementations to go with that model it seems like with created and modified could always be derived from started and stopped (or vice-versa).

carolynvs: Are you suggesting that all implementations would have to make a single result document per "event", e.g. the operation is pending is one result, when the execution starts that's another result, etc. So the result document wouldn't be mutable either?

jlegrone: Yes, this would also prevent runtimes from overwriting potentially important metadata for debugging purposes, like exactly when an output file was generated or the `result.message` for some intermediate step in the claim's lifecycle.

Runtimes could choose to persist only the first and last observed results, which would end up being pretty much the same as using a single mutable result and continuously updating its fields.

carolynvs: Gotcha! 👍 That seems fine with me. This pushes the claim spec much closer to laying out an implementation but honestly I think that is a good thing.

carolynvs:
> I'd be up for adding an id attribute to claims

Great! I know that wasn't palatable before. Let me rework the PR today to address your feedback.

carolynvs:
@technosophos @jlegrone I reverted my last commit since it caused unanticipated problems with DataDog's implementation.

Please take a **close** look as I am altering or perhaps clarifying terms and fields that different implementations have had different interpretations for. 😀 

---

* Installation refers to the installation of the bundle
* Claim records the execution of an action against an installation
* Claims have an id that results reference
* An installation has _many_ claims with these new definitions.
* Claims and results are immutable. Each new action against an
installation generates a new claim. Each new status transition of an
operation could yield a new result (depending on the implementation).
* Clarify that the last modifying claim is retained (same as before)
unless the implemenation wants more history. The last result is retained
for a retained claim (again unless an implemenation wants more history).

Depending on your previous interpretation of the spec, this may change what claim.created means to you. The meaning has hopefully been clarified to the created date of the claim (see updated definition above), not the installation.

Both Porter and Duffle would update to keep more history. A more simple implementation could choose to keep a single file around per installation with the last claim revision and the associated result. It would lose the built in answer to "When was the entire installation first created?" (see above) but could put that into claim.custom if it cared to know, or could keep around the first revision, and along with the last.

jlegrone:
I think this closes https://github.com/cnabio/cnab-spec/issues/328?

400-claims.md
```diff
@@ -128,64 +130,90 @@ The parameter data stored in a claim data is _the resolved key/value pairs_ that
...
+}
+```
+Source: [400.01-claim-result.json](examples/400.01-claim-result.json)
 
-The claim `result` provides a list of output names generated by the operation. A tool may choose to request the contents of any outputs to persist them but is not required to do so.
+The fields above are defined as follows:
+
+- `claimId` (REQUIRED): ID of the claim that generated this result.
+- `custom` (OPTIONAL): A section for custom extension data applicable to a given runtime.
+- `id` (REQUIRED): An [ULID](https://github.com/ulid/spec) identifier for the result.
```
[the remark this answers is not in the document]
jlegrone: We talked about this offline. My understanding is that `id` is meant to be used in the go-cnab storage driver. This makes sense; thanks for indulging my questions. 😄

400-claims.md
```diff
@@ -46,71 +46,73 @@ This is done so that implementors can standardize on a way of relating an instal
...
-  },
   "revision": "01CP6XM0KVB9V1BQDZ9NK8VP29"
 }
 ```
 Source: [400.01-claim.json](examples/400.01-claim.json)
 
 The fields above are defined as follows:
 
-- `action` (REQUIRED): The name of the action that last caused this claim to be updated. This may be any of the built-in actions (`install`, `upgrade`, `uninstall`) as well as any custom actions as defined in the bundle descriptor. The special name `unknown` MAY be used in the case where the CNAB Runtime cannot determine the action name of a claim.
+- `action` (REQUIRED): The name of the action executed against the installation. This may be any of the built-in actions (`install`, `upgrade`, `uninstall`) as well as any custom actions as defined in the bundle descriptor. The special name `unknown` MAY be used in the case where the CNAB Runtime cannot determine the action name of a claim.
```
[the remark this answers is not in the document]
carolynvs: `unknown` has been listed in the spec as a reserved state I think since the beginning? I'm with you that I have no idea what a runtime would do with what is essentially an invalid state though. I know Porter would just end up not running anything and fail. It's not terribly useful to make it an explicit state. 😀

400-claims.md
```diff
@@ -46,71 +46,73 @@ This is done so that implementors can standardize on a way of relating an instal
...
 - `bundle` (REQUIRED): The bundle, as defined in [the Bundle Definition](101-bundle-json.md).
-- `created` (OPTIONAL): A timestamp indicating when this release claim was first created. This MUST not be changed after initial creation.
-- `bundleReference` (OPTIONAL): A canonical reference to the bundle used in the last action. This bundle reference SHOULD be digested to identify a specific version of the referenced bundle.
+- `created` (REQUIRED): A timestamp indicating when this _claim_ was created.
+- `bundleReference` (OPTIONAL): A canonical reference to the bundle used in the action. This bundle reference SHOULD be digested to identify a specific version of the referenced bundle.
```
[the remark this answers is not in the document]
carolynvs: Yes, this is out of scope for this PR so please open up a separate issue. 👍

400-claims.md
```diff
@@ -128,64 +130,90 @@ The parameter data stored in a claim data is _the resolved key/value pairs_ that
...
 
-The claim `result` provides a list of output names generated by the operation. A tool may choose to request the contents of any outputs to persist them but is not required to do so.
+The fields above are defined as follows:
+
+- `claimId` (REQUIRED): ID of the claim that generated this result.
+- `custom` (OPTIONAL): A section for custom extension data applicable to a given runtime.
+- `id` (REQUIRED): An [ULID](https://github.com/ulid/spec) identifier for the result.
+- `created` (REQUIRED): A timestamp indicating when this result was created.
+- `message` (OPTIONAL): A human-readable string that communicates the outcome. Error messages MAY be included in `failed` conditions.
+- `outputs` (OPTIONAL): Outputs generated by the operation. It is a map from the output names to metadata about the output. The  output value is not stored in the claim result. See the [Outputs](#outputs) section for details. If this field is not present, it may be assumed that no outputs were generated as a result of the operation.
```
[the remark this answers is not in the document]
carolynvs: I just added contentDigest as an optional piece of metadata that may be included.

carolynvs:
This is gonna be an epic squash

womens squash tournament

Issue #346, opened 2020-03-05
Proposal: Enable versioned schema validation


vdice:
# Goal

CNAB Implementations have a need to validate against specific versions of CNAB schemata

# Proposed Requirements

The CNAB Spec authority (cnabio/cnab-spec) should:
  - Produce schema artifacts for each specific schema version and persist these in a publicly-available manner (Note that, at time of writing, the schema versions currently visible/served via e.g. https://cnab.io/v1/bundle.schema.json always represent the latest/`master` branch versions)

Implementations should be able to fetch the version(s) of schemata they implement and:
  - May use the schemata to validate implementations at build/test-time
  - May include the schemata into their binaries/library code for runtime validation or downstream consumers

# Proposed Implementations

Proposed implementation for cnabio/cnab-spec to achieve the requirement(s) above:

  - Issue a git tag for each schema version
    - e.g. `cnab-core-v1.0.0`, `cnab-claim-v1.0.0-DRAFT`, `cnab-dependencies-v1.0.0-DRAFT`
      - For schemata still in draft, I propose we cut additional tags for implementation-use only (these aren't considered official schema versions).  In doing so, we enable implementations to incrementally update to the latest version of the working draft schema.
        - e.g. `cnab-claim-v1.0.0-DRAFT+8a4242d` (short git SHA points to exact commit)

  - Configure CI to produce artifacts (i.e. all schema.json files present at time of tag) based on the pushed git tag.  A couple options (not mutually exclusive, if we'd like both):

    - Push to blob store
      - Upload the artifacts (schema.json files) to a location with incorporation of the git tag/version in the name
        - e.g. `cnab-core-v1.0.0/bundle.schema.json`
      - Enable anonymous users to download this artifact
        - Perhaps ideally via a custom domain, e.g. `https://cdn.cnab.io/cnab-core-v1.0.0/bundle.schema.json`.  This also would/could enable web view.
      - An additional benefit: The [cnab.io website](https://github.com/cnabio/cnab.io) should be able to serve all of the same cdn-hosted assets via a single Netlify redirect entry, e.g. `/schema/* https://cdn.cnab.io/schema/:splat`.  Resulting in, say, `https://cnab.io/schema/cnab-core-v1.0.0/bundle.schema.json`

    - Push as OCI artifact to compatible registry (with anonymous access)
      - Can utilize oras (https://github.com/deislabs/oras) to push artifacts to a compatible registry (ACR has compatibility)
[...]

PR #350, opened 2020-03-10
feat(ci): publish versioned schema

vdice:
* adds CI for publishing schema on git tag pushes
* adds entry on where to find published schema in README
* refactors brigade.js with latest brigade-utils helpers

Contributes towards https://github.com/cnabio/cnab-spec/issues/346
Depends on https://github.com/cnabio/cnab.io/pull/32 (merged)

TODO (after approval):

 - [ ] update PR check from `validation` to `cnab-spec-validate`:

Screen Shot 2020-03-10 at 1 52 11 PM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
-BASE_DIR          := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
-VALIDATOR_IMG     := deislabs/cnab-spec.ajv
+VERSION       ?= $(shell git describe --tags 2> /dev/null || echo v0)
+BASE_DIR      := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+VALIDATOR_IMG := deislabs/cnab-spec.ajv
 
 .PHONY: build-validator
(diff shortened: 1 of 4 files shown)


vdice:
@technosophos I just realized the following:  I had centered CI around git tags but I now recall that we have a precedent of pushing spec versions to separate branches.  Can we change the process to instead cut tags for each future spec version?  Or should I adjust this process to trigger on branch creation?  (May be trickier to ignore potential non-version-related branches, etc.)

vdice:
> @technosophos I just realized the following: I had centered CI around git tags but...

Discussion transitioned to https://github.com/cnabio/cnab-spec/issues/351

technosophos:
I think we can publish a tag for 1.0.1 if we need to get this working.

vdice:
> I think we can publish a tag for 1.0.1 if we need to get this working.

It looks like a tag can't be pushed with the same name of a branch.  I've manually uploaded the schema artifacts from the `cnab-core-1.0.1` branch as a stop-gap 👍 .

Issue #351, opened 2020-03-11
Proposal: Versioning Strategy (Git + JSON Schema)


vdice:
In this repository, we've yet to fully settle on a given versioning strategy for the JSON Schema files created here.  Perhaps the main source of friction/disconnect is aligning a Git versioning strategy with a proper JSON Schema versioning strategy.

So far, we've tried a tag for the first official release: [cnab-core-1.0](https://github.com/cnabio/cnab-spec/releases/tag/cnab-core-1.0)

For the next CNAB Core release, we used a branching model: [cnab-core-1.0.1](https://github.com/cnabio/cnab-spec/tree/cnab-core-1.0.1).  The intention here was to support the approved CNAB Governance mechanism for amending typographical errata without necessarily cutting a new patch release.

However, whilst exploring options for producing artifacts from this repository at pinned commits (tags or branches, etc.) in https://github.com/cnabio/cnab-spec/issues/346 and its current implementation PR (https://github.com/cnabio/cnab-spec/pull/350), it became clearer that Git tags may be the more convenient method for pinning spec/schema versions (and enabling CI to publish assets from these commits).

I may be missing some other pertinent points and may not be relaying all related issues, but I'm inclined to propose the following:

1. On each agreed spec version increment, we cut a Git tag, e.g. `cnab-core-1.1.0`, `cnab-claim-1.0.0` and even draft versions at certain commits, such as `cnab-claim-1.0.0-DRAFT+abc1234`, where `abc1234` represents the short Git SHA associated with the tagged commit.
    - On these events, CI will publish all schema files located in the `schema` sub-directory according to the tag, accessible via e.g. `https://cnab.io/schema/cnab-core-1.1.0/bundle.schema.json`
    - Implementing tools would then set the schema version on their corresponding implementations based on these tags.  For example, a tool implementing a bundle at tag `cnab-core-1.1.0` would set `"schemaVersion":"1.1.0"` for the produced `bundle.json` file.  Or, a tool implementing a (working draft) claim at tag `cnab-claim-1.0.0-DRAFT+abc1234` would set `"schemaVersion":"1.0.0-DRAFT+abc1234"` for the produced `claim.json` file.

[...]

PR #353, opened 2020-03-12
Remove outdated reference to `bundle.cnab`

radu-matei:
The 100 section apparently still had two references to a signed bundle being `bundle.cnab`.

This PR _just_ removes those two references, but I suspect we could also add a note in the _History_ section describing the changes.

The main question is whether this can be considered an errata change (see #351).

This PR is opened against the 1.0.1 branch - is this the right approach?
How should `master` be updated?

cc @chris-crone @technosophos


vdice:
> This PR is opened against the 1.0.1 branch - is this the right approach?
How should master be updated?

Perhaps we either cherry-pick from the 1.0.1 branch... or start via a PR against master and then cherry-pick into the 1.0.1 branch (reverse)?

carolynvs:
+1 for this being treated as a patch since this wasn't implementable

technosophos:
@vdice are you okay with me merging this onto `master`?

vdice:
@technosophos 👍

Issue #352, opened 2020-03-12
Optional invocation images


glyn:
# Optional Invocation Images

In environments, notably Kubernetes, where applications define their installations declaratively, the presence of an invocation image in a bundle is seen by some as a potential security exposure (similar to ActiveX). This is likely to limit the adoption of CNAB in those environments.

The invocation image should be made optional. A bundle without an invocation image can be installed only by runtimes that support the bundle's metadata and/or extension(s).

This will make CNAB more broadly applicable, notably within the Kubernetes community. Features of the CNAB spec other than invocation images, including image relocation, air gap support, registry representation, and supply chain security, can then be exploited without the necessity of using invocation images.

# A Spectrum of Experiences

This spec change enables CNAB to support the following experiences:

* A purely declarative install mechanism. All necessary artifacts (with metadata about those artifacts) would be included in the bundle. External tools would be required to interpret and use those artifacts to do... something. It could be installing something, but it could also be other things over time.

* Move the install tool chain into a container (or set of containers) that are left as a suggestion (e.g. because the invocation image is omitted) in the metadata on how to use the artifacts. It would be up to the user to use (or not use) those containers.

* Have everything bundled into the install image much as it is today. That image would be opaque and must be used to install the application as some of the required artifacts may be baked into that image.

# Compatibility

This is a backward compatible spec change since all existing bundles are valid in the updated spec. However, bundles without invocation images are invalid in the previous version of the spec. This change is therefore proposed for CNAB v1.1.

Since there is no standard way to perform actions on a bundle without an invocation image, the reference implementation needs updating to fail when asked to perform an action on such a bundle, but to allow such bundles when not performing an action.

# Context

* Declarative Installation Support ([issue 285](https://github.com/cnabio/cnab-spec/issues/285))
* [VMware discussions](https://github.com/cnabio/cnab-spec/issues/285#issuecomment-591724062) about which tool chain to standardise and support for the Tanzu product line. Also the [`sheaf`](https://github.com/bryanl/sheaf) prototype.
* Extension for Declarative Kubernetes Installers ([issue 338](https://github.com/cnabio/cnab-spec/issues/338))

glyn:
After discussion in the [CNAB Community Call](https://hackmd.io/TfLYtyRnS8Kfx4ekgLwWLg) on 18 March 202, it appears that making invocation images optional bifurcates the spec and makes interoperability more difficult. Closing.

PR #358, opened 2020-03-16
Make invocation images optional

glyn:
Fixes https://github.com/cnabio/cnab-spec/issues/352

### DIFF
--- a/000-index.md
+++ b/000-index.md
@@ -5,9 +5,9 @@ weight: 001
 
 # CNAB Specifications
 
-1. [Cloud Native Application Bundle Core 1.0.0 (CNAB1)](100-CNAB.md)
+1. [Cloud Native Application Bundle Core 1.1.0 (CNAB1.1)](100-CNAB.md)
     - [The bundle.json File](101-bundle-json.md)
-    - [The Invocation Image Format](102-invocation-image.md)
+    - [Invocation Image Format](102-invocation-image.md)
     - [The Bundle Runtime](103-bundle-runtime.md)
     - [Bundle Formats (Thick and Thin)](104-bundle-formats.md)
 2. [Cloud Native Application Bundle Registry 1.0.0 (CNAB-Reg)](200-CNAB-registries.md)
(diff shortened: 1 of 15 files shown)


radu-matei:
This PR does make invocation images optional _syntactically_ in the specification text, but it does _not_ have any details as to how potential runtimes should handle things like parameters, credentials, or outputs. 

As far as I can tell, this is the only addition to the runtime spec:

> If an attempt is made to execute the run tool of a bundle with no invocation image, the runtime MUST exit with an error. A runtime MAY support CNAB actions on bundles with no invocation image (e.g. by handing the action itself and not attempting to execute the bundle's run tool).

glyn:
> This PR does make invocation images optional _syntactically_ in the specification text, but it does _not_ have any details as to how potential runtimes should handle things like parameters, credentials, or outputs.

Yes, I need to add some text about that, thanks.

glyn:
@radu-matei I've added some text about parameters, credentials, and outputs. I've also provided a non-normative summary in 806-declarative-installation-support.md.

carolynvs:
@glyn This pull request was originally scoped to "_make invocation images optional_". However the latest changes make the majority of the CNAB specification optional for bundles that don't use invocation images.

A bundle that doesn't use invocation images also doesn't use: parameters, credentials and outputs. This feels unnecessary. I would prefer that any proposal includes a solution that keeps the existing (useful) features of the CNAB spec and sticks to the original scope of only making invocation images optional.

glyn:
> @glyn This pull request was originally scoped to "_make invocation images optional_". However the latest changes make the majority of the CNAB specification optional for bundles that don't use invocation images.

What could be normative and worth saying about parameters, credentials, and outputs for bundles with no invocation image? Maybe I've misunderstood the intention behind those features and how closely they are tied to invocation images.

> 
> A bundle that doesn't use invocation images also doesn't use: parameters, credentials and outputs. This feels unnecessary. I would prefer that any proposal includes a solution that keeps the existing (useful) features of the CNAB spec and sticks to the original scope of only making invocation images optional.

I understand. Please note that I deliberately used the word SHOULD in sentences such as:
> If a bundle has no invocation images, the parameters section SHOULD be omitted.

The idea was to allow parameters, credentials, and outputs to be used if and only if the runtime(s) that the bundle is targetting can handle them. I wonder if there's a better way of expressing this. How about, for example:
> If a bundle has no invocation images, the parameters section MAY be omitted.

Any thoughts greatly appreciated.

glyn:
After discussion in the [CNAB Community Call](https://hackmd.io/TfLYtyRnS8Kfx4ekgLwWLg) on 18 March 202, it appears that making invocation images optional bifurcates the spec and makes interoperability more difficult. Closing.

PR #364, opened 2020-04-15
Clarify how to verify & inspect in-toto metadata

trishankatdatadog:


303-verification-workflows.md
```diff
@@ -23,7 +23,7 @@ In order to download and verify a bundle, a compliant runtime MUST take the foll
...
+1. Call the in-toto runtime to [verify and inspect](https://github.com/in-toto/docs/blob/e9806a000c32dea73f6044a140386f601c7d4e18/in-toto-spec.md#52-verifying-the-final-product) the bundle using the downloaded root layout, public keys for the root layout, link metadata, and the bundle itself. The bundle runtime MAY run in-toto directly on the operating system, in a default container image provided by the bundle runtime, or in a custom container image provided by the user. It is up to the bundle developers to ensure that in-toto inspections are as portable as possible. If the verification and inspection of the bundle fails for any reason, the runtime MUST discard the downloaded bundle, report the error, and stop execution.  Otherwise, it MUST halt, and return the verified bundle to the user.
```
silvin-lubecki, line 26: nit: extra space before `Otherwise`

silvin-lubecki: Question:
> Otherwise, it MUST halt, and return the verified bundle to the user.

Do we really need to specify that the process needs to halt? 🤔 
Why does it need to return the verified bundle? Did the in-toto runtime mutate it? Otherwise it should be the same as the downloaded one?

carolynvs: I'm confused by this sentence as well. Does "otherwise" refer to if the validation passes? I have the same question as Silvin.

trishankatdatadog: Thanks for comments! Please see if it's clearer now?

PR #375, opened 2020-06-15
Move security known implementations document to the non-normative section

radu-matei:


radu-matei:
Oh yes, actually pushing all the changes is probably helpful...thanks!

806-security-known-implementations.md
```diff
@@ -1,9 +1,9 @@
 ---
-title: CNAB Security: Known implementations
-weight: 304
+title: Known implementations for CNAB Security
+weight: 805
```
trishankatdatadog, line 3:
```suggestion
weight: 806
```

radu-matei: Good catch!

PR #383, opened 2020-08-18
500-CNAB-dependencies-sequence

### LINKED ISSUE #361: Change proposal: `500-CNAB-dependencies` - section [Depend on a named bundle]
### Background

Since the nested dependencies are not supported yet. So we decided to create multiple dependent bundles, and we want them to execute in a certain orders. But on each run we get a random execution at Runtime. Here what we get:

##### Declared

```shell
MYDEP-core
MYDEP-s3
MYDEP-db
MYDEP-experiments
```

##### Executed

```shell
MYDEP-experiments
MYDEP-s3
MYDEP-core
MYDEP-db
```
### Investigation

After a back-and-forth with ***Porter Authors*** @carolynvs  and @vdice, we think it would be more practical to have an array of dependencies and introduce a `name` field to insure bundle identification and maintain the functionality of templating values, e.g. `{{ bundle.dependencies.mysql.output.foo }}`.

```yaml
dependencies:
  - name: mysql
    tag: myregistry/mysql:v0.1.0
  - name: wordpress
    tag: myregistry/wordpress:v0.1.0
```

> Related issue: https://github.com/deislabs/porter/issues/969

### Proposal

Change the `requires` listing to an array, an individual listing now has a name field. Thus, Runtimes are expected to process listings in order of appearance. Thus we would have:

```json
{
  "custom":{
    "dependencies":{
      "requires":[
        {
          "name":"storage",
          "bundle":"somecloud/blob-storage"
        },
        {
          "name":"mysql",
          "bundle":"somecloud/mysql",
          "version":{
            "prereleases":true,
            "ranges":[
              "5.7.x"
            ]
          }
        }
      ]
    },
    "name":"wordpress"
  }
}
```

MChorfa:
# What does this change
Add sequence definition to CNAB-dependencies in order to enable ordering behavior

# What issue does it fix
Closes #361

### DIFF
--- a/500-CNAB-dependencies.md
+++ b/500-CNAB-dependencies.md
@@ -70,8 +70,10 @@ MUST only be used for this CNAB Dependencies Specification.
 
 ### sequence
 
-The `sequence` array defines the criteria for sequencing bundles in a pre-defined order. The sequence for
-each occurrence is a way for the bundle to reference the order of a dependency.
+The `sequence` array is a pre-defined sequencing of the dependant bundles.
+If the `sequence` array is non-empty, the runtime MUST execute each dependency according to this pre-defined order
+unless the runtime is capable of resolving the dependency graph to determine an execution order itself.
+The `sequence` for each occurrence is a way for the bundle to reference the order of a dependency.
 
 ### requires
 
--- a/500-CNAB-dependencies.md
+++ b/500-CNAB-dependencies.md
@@ -70,7 +70,7 @@ MUST only be used for this CNAB Dependencies Specification.
 
 ### sequence
 
-The `sequence` array defines the criteria for the sequencing bundles in a pre-defined order. The sequence for
+The `sequence` array defines the criteria for sequencing bundles in a pre-defined order. The sequence for
 each occurrence is a way for the bundle to reference the order of a dependency.
 
 ### requires
--- a/500-CNAB-dependencies.md
+++ b/500-CNAB-dependencies.md
@@ -31,13 +31,14 @@ There are two cases for how a bundle may need to depend upon another bundle:
 
 ## Depend on a named bundle
 
-The bundle depends on a specific named bundle that is known in advance. It is 
+The bundle depends on a specific named bundle that is known in advance. It is
 stored in the custom extensions section of the bundle.
 
 ```json
 {
   "custom": {
     "dependencies": {
+      "sequence": ["storage", "mysql"],
       "requires": {
         "storage": {
           "bundle": "somecloud/blob-storage"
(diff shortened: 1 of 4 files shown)


carolynvs:
@MChorfa This looks great so far! 👍 Once it's ready for review, remove the WIP and I'm eager to approve.

carolynvs (approved):
LGTM

500-CNAB-dependencies.md
```diff
@@ -64,9 +65,14 @@ This section is a placeholder and will be completed in a follow-up pull request.
 This specification introduces a `dependencies` object in the bundle.json
 that defines metadata necessary to specify a dependency.
 
-The entry `dependencies` in the custom extension map, `custom`, is reserved and
+The entry `dependencies` in the custom extension object, `custom`, is reserved and
 MUST only be used for this CNAB Dependencies Specification.
 
+### sequence
+
+The `sequence` array defines the criteria for sequencing bundles in a pre-defined order. The sequence for
```
vdice, line 73: Do we want to entertain stricter verbiage here?  Something like, 'If the sequence array is non-empty, the runtime MUST execute each dependency according to this pre-defined order'?

And/or, if we don't wish to make it a MUST, perhaps we should update to specify SHOULD?

MChorfa: Yes, I would agree with `SHOULD` to give flexibility. How about : 
> The `sequence` array defines the criteria for sequencing dependant bundles, if the `sequence` array is non-empty, the runtime SHOULD execute each dependency according to this pre-defined order. 
The sequence for each occurrence is a way for the bundle to reference the order of a dependency.
Also, do we have to indicate that number of the `sequences` has to mutch the number of `requires` to perform the ordering? Otherwise, we handle only the `requires`.

carolynvs: I am concerned about the looser SHOULD. Is that intended to allow a runtime to resolve the dependency graph and determine the execution order of the dependencies itself? Or to allow a runtime to opt out of executing dependencies in any reliable order at all?

We can't allow the spec to say it's okay in the second case to ignore the author and run the dependencies like they are today, unordered. An author needs to be able to rely on the dependencies being run in the proper order, otherwise it can unexpectedly fail while executing because the order is incorrect.

For the first, where we want to allow for smarter runtimes but disallow ignoring ordering entirely, how about something like this?

> The sequence array is a pre-defined sequencing of the dependant bundles. If the sequence array is non-empty, the runtime MUST execute each dependency according to this pre-defined order unless the runtime is capable of resolving the dependency graph to determine an execution order itself. The sequence for each occurrence is a way for the bundle to reference the order of a dependency.

MChorfa: Yesss ... I love it :)

carolynvs:
This is amazing timing. I just released Porter v0.29.0 this morning with your implementation of this spec change. 🎉

Issue #287, opened 2019-10-14
Clarify contentDigest meaning for docker/OCI images


glyn:
Let's start with some definitions (based on the ["OCI Image Format Specification"](https://github.com/opencontainers/image-spec/blob/master/spec.md)).

Docker and OCI images have two types of digest: a _repo digest_ and an _image id_. A repo digest is the SHA-256 digest of the _compressed_ image manifest. Since compression depends on the implementation of the registry used to store the image, the repo digest doesn't logically exist until the image has been pushed. An image id, on the other hand, is the SHA-256 digest of the _uncompressed_ image configuration, which is independent of the registry implementation.

Both these digests are content addresses of an image in the sense that each uniquely identifies the content (modulo SHA-256 collisions). Note that the [docker registry spec](https://docs.docker.com/registry/spec/api/#content-digests) refers to the repo digest as a "content digest".

The [CNAB spec](https://github.com/deislabs/cnab-spec/blob/master/101-bundle-json.md) defines the `contentDigest` fields in bundle.json as follows, firstly for invocation images:
> The contentDigest field MUST contain a digest, in OCI format, to be used to compute the integrity of the image. The calculation of how the image matches the contentDigest is dependent upon image type. (OCI, for example, uses a Merkle tree while VM images are checksums). During bundle development, it may be ideal to omit the contentDigest field and/or skip validation. Once a bundle is ready to be transmitted as a thick or thin bundle, it must have a contentDigest field. If a contentDigest field is present, a runtime MUST validate the image digest prior to executing an action. If the contentDigest is not present, the runtime SHOULD report an error so the user is aware that there is no contentDigest provided. Runtimes MAY allow users to override this behavior and perform actions on bundles that do not have contentDigest values populated.

and then for images other than invocation images:
> contentDigest: MUST contain a digest of the contents of the image, in OCI format, to be used to compute the integrity of the image. The calculation of how the image matches the contentDigest is dependent upon image type. (OCI, for example, uses a Merkle tree while VM images use checksums.)

Since both repo digests and image ids are roots of Merkle trees, the CNAB spec doesn't actually prescribe whether repo digest or image id (or indeed some other Merkle tree root digest!) should be used for `contentDigest` fields of docker/OCI images. This needs clarifying so that CNAB runtimes know how to validate these fields.

trishankatdatadog:
Great question! I think _image id_ is the better idea, since it's registry-independent... @jlegrone

jeremyrickard:
[...]

PR #384, opened 2020-08-20
docs(101-bundle-json.md): add clarification around contentDigest value

vdice:
* Clarifies which type of digest must be used for the contentDigest field of an invocation or bundle image

Proposed fix for https://github.com/cnabio/cnab-spec/issues/287

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -315,7 +315,7 @@ The `imageType` field MUST describe the format of the image. The list of formats
 The `image` field MUST give a path-like or URI-like representation of the location of the image. It is REQUIRED. The expectation is that an installer should be able to locate the image (given the image type) without additional information.
 
 The `contentDigest` field MUST contain a digest, in [OCI format](https://github.com/opencontainers/image-spec/blob/master/descriptor.md#digests), to be used to compute the integrity of the image. The calculation of how the image matches the `contentDigest` is dependent upon image type. (OCI, for example, uses a Merkle tree while VM images are checksums).
-  - If the `imageType` is `oci` or `docker`, the value for this field MUST be the registry-compressed digest (known as a repo digest, [image digest](https://docs.docker.com/engine/reference/commandline/images/#list-image-digests) and/or manifest digest) as opposed to the registry-independent image ID.
+  - If the `imageType` is `oci` or `docker`, the value for this field MUST be the registry-compressed digest (known as a repo digest, [image digest][image-digest] and/or manifest digest) as opposed to the registry-independent image ID.
   - During bundle development, it may be ideal to omit the `contentDigest` field and/or skip validation. Once a bundle is ready to be transmitted as a thick or thin bundle, it MUST have a `contentDigest` field.
   - If a `contentDigest` field is present, a runtime MUST validate the image digest prior to executing an action
   - If the `contentDigest` is not present, the runtime SHOULD report an error so the user is aware that there is no `contentDigest` provided.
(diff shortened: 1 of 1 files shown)


101-bundle-json.md
```diff
@@ -314,7 +314,12 @@ The `imageType` field MUST describe the format of the image. The list of formats
...
+The `contentDigest` field MUST contain a digest, in [OCI format](https://github.com/opencontainers/image-spec/blob/master/descriptor.md#digests), to be used to compute the integrity of the image. The calculation of how the image matches the `contentDigest` is dependent upon image type. (OCI, for example, uses a Merkle tree while VM images are checksums).
+  - The value for this field MUST be the registry-compressed digest (known as a repo digest, [image digest](https://docs.docker.com/engine/reference/commandline/images/#list-image-digests) and/or manifest digest) as opposed to the registry-independent image ID.
```
carolynvs, line 318: This should only be required when the image type is oci.

101-bundle-json.md
```diff
@@ -353,7 +358,7 @@ Fields:
...
   - `imageType`: The `imageType` field MUST describe the format of the image. The list of formats is open-ended, but any CNAB-compliant system MUST implement `docker` and `oci`. The default is `oci`.
   - `image`: The REQUIRED `image` field provides a valid reference for the image. Note that SHOULD be a CAS SHA, as in the example above, not a version tag.
-  - `contentDigest`: MUST contain a digest of the contents of the image, in [OCI format](https://github.com/opencontainers/image-spec/blob/master/descriptor.md#digests), to be used to compute the integrity of the image. The calculation of how the image matches the contentDigest is dependent upon image type. (OCI, for example, uses a Merkle tree while VM images use checksums.)
+  - `contentDigest`: MUST contain a registry-compressed digest of the contents of the image, in [OCI format](https://github.com/opencontainers/image-spec/blob/master/descriptor.md#digests), to be used to compute the integrity of the image. The calculation of how the image matches the contentDigest is dependent upon image type. (OCI, for example, uses a Merkle tree while VM images use checksums.)
```
carolynvs, line 361: Same for here, the type of digest is dependent on the image type.

SteveLasker:
Curious why the digest is required to be consistent with the tag.
While deploying unique tags is a best practice to have consistent deployments across a scale unit, locking to the digest blocks the "break glass" servicing capability. I get that some use [docker-lock](https://github.com/safe-waters/docker-lock) type rules until we have a better signing, or standard tag locking capability. Just suggesting this shouldn't be a hard requirement. 

As a minor point, some of the references refer to  `"image":"example/microservice:1.2.3"`
This is a bit confusing. Is this the default docker hub behavior where no registry = docker.io? Or, does the bundle support finding the image in the registry where the CNAB bundle was pulled from?
There's some later examples of: `"image": "example.com/example/vote-frontend@sha256:aca460afa270d4c527981ef9ca4989346c56cf9b20217dcea37df1ece8120685",`
Would suggest fully qualified names, if that's what it means, or better yet, support reference artifacts to the registry where the CNAB was pulled enabling better transportability.

For the image reference, the spec states:
> The following OPTIONAL fields MAY be attached to an invocation image:
> size: The image size in bytes. Implementations SHOULD verify this when a bundle is packaged as a thick bundle, and MAY verify it when the image is part of a thin bundle.

Would suggest requiring the full [OCI Descriptor](https://github.com/opencontainers/image-spec/blob/master/descriptor.md#properties) information, where `digest`, `mediaType` and `size` are required.

chris-crone (approved):
LGTM

This is a clarification but it would, in theory, break runtimes that use a different `contentDigest` definition so I believe that rules this out as a simple errata. We'll need to bump the spec version.

vdice:
@SteveLasker good points, all.  Please feel free to create follow-up issues for spec amendments/proposals mentioned.  However, the scope of this changeset is simply to clarify the expected value of the `contentDigest` field when the invocation image type is docker/oci.

trishankatdatadog (commented):
Hmm, interesting. Would this break Signy's verification behaviour, @radu-matei? I suppose not unless you consider copying images between different registries?

technosophos (approved):
LGTM. Much clearer and precise this way.

chris-crone:
> Hmm, interesting. Would this break Signy's verification behaviour, @radu-matei? I suppose not unless you consider copying images between different registries?

I remember having discussions about this with @radu-matei so I think he likely implemented Signy with this definition in mind

silvin-lubecki (approved):
LGTM 👍

PR #386, opened 2020-09-16
feat(claims): update claims spec and result schema

vdice:
Whilst re-reviewing the current state of the Claims Spec doc, I found some areas that appear to need updating.

Also, added the missing `custom` field to the claim result schema.

This is in anticipation of finalizing the Claims Spec for its 1.0.0 release.

### DIFF
--- a/400-claims.md
+++ b/400-claims.md
@@ -18,7 +18,7 @@ A _host environment_ is an environment, possibly shared between multiple CNAB ru
 
 The word _claim_ was chosen because it represents the relationship between a certain CNAB host environment and the resources that were created by a CNAB runtime in that host environment. In this sense, an environment takes responsibility for those resources if and only if it can _claim_ them. A claim is an _external assertion of ownership_. That is, the claim itself is not "installed" into the host environment. It is stored separately, possibly in an entirely different location.
 
-The claims system is designed to satisfy the requirements of the [Bundle Runtime specification](103-bundle-runtime.md) regarding the tracking `CNAB_REVISION` and `CNAB_LAST_REVISION`. It also provides a description of how the state of a bundle installation may be represented.
+The claims system is designed to satisfy the requirements of the [Bundle Runtime specification](103-bundle-runtime.md) regarding the tracking of `CNAB_REVISION` and `CNAB_LAST_REVISION`. It also provides a description of how the state of a bundle installation may be represented.
 
 This specification uses the terminology defined in [the CNAB Core specification](100-CNAB.md).
 
@@ -231,7 +231,7 @@ WfK01YbCWioNVGk=
 
 ### How the Claim is Used
 
-The claim is used to inform any CNAB tooling about how to address a particular installation. For example, given the claim record, a package manager that implements CNAB should be able to:
+The claim is used to inform any CNAB tooling about how to address a particular installation. For example, given the claim record, a runtime that implements the CNAB Claims spec should be able to:
 
 - List the _names_ of the installations, given a _bundle name_
 - Given an installation's _name_, return the _bundle info_ that is installed under that name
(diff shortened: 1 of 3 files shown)


400-claims.md
```diff
@@ -105,7 +105,7 @@ The fields above are defined as follows:
...
 - `installation` (REQUIRED): The name of the _installation_. This can be automatically generated, though humans may need to interact with it. It MUST be unique within the installation environment, though that constraint MUST be imposed externally. Elsewhere, this field is referenced as the _installation name_. The format of this field must follow the same format used for the `installation` field in the [bundle.json file specification](101-bundle-json.md#the-bundlejson-file).
-- `parameters` (OPTIONAL): Key/value pairs that were passed in during the operation. These are stored so that the operation can be re-run. Some implementations MAY choose not to store these for security or portability reasons. However, there are some caveats. See the [Parameters](#parameters) section below for details.
+- `parameters` (OPTIONAL): Key/value pairs that were passed in during the operation. These are stored so that the operation can be re-run. See the [Parameters](#parameters) section below for more details.
```
vdice, line 108: Here and in the `Parameters` section below, I've updated verbiage according to the [section at the end of this document](https://github.com/cnabio/cnab-spec/blob/master/400-claims.md#credentials-parameters-and-claims).

400-claims.md
```diff
@@ -231,24 +231,23 @@ WfK01YbCWioNVGk=
...
+The claim is used to inform any CNAB tooling about how to address a particular installation. For example, given the claim record, a runtime that implements the CNAB Claims spec should be able to:
 
 - List the _names_ of the installations, given a _bundle name_
 - Given an installation's _name_, return the _bundle info_ that is installed under that name
 - Given an installation _name_ and a _bundle_, generate a _bundle info_.
   - This is accompanied by running the `install` path in the bundle
 - Given an installation's _name_, replace the _bundle info_ with updated _bundle info_, and update the revision with a new ULID, and the modified timestamp with the current time. This is an upgrade operation.
   - This is accompanied by running the `upgrade` path in the bundle
-- Given an installation's name, mark the claim as deleted.
+- Given an installation's name, mark the installation as deleted.
```
vdice, line 242: We might actually want to change `deleted` to `uninstalled`.

carolynvs: Agreed 👍

400-claims.md
```diff
@@ -269,7 +268,7 @@ Parameters declared with an `env` key in the `destination` object MUST have thei
 
 The invocation image may benefit from accessing the claim. Consequently, a claim MUST be attached to the invocation image when the invocation image is started.
 
-The claim MUST be mounted at the path `/cnab/claim.json` inside of the bundle. The version of this claim that is to be mounted is the _version prior to this operation beginning_. In other words, when a bundle is installed, it creates the original installation claim. On the first upgrade, the claim describing the _installation_ is located at `/cnab/claim.json`. This allows the invocation image to inspect the former state and compare it to the desired new state.
+The claim MUST be mounted at the path `/cnab/claim.json` inside of the bundle. The version of this claim that is to be mounted is the _version representing the current operation_. In other words, when a bundle is installed, the runtime creates the original installation claim and passes this in. On the first upgrade, the claim describing the _upgrade_ operation is located at `/cnab/claim.json`.
```
carolynvs, line 271: I am very happy to fix this to match what we have always implemented. 👍

schema/claim-result.schema.json
```diff
@@ -32,6 +32,9 @@
       "description": "The date created, as an ISO-8601 Extended Format date string, as specified in the ECMAScript standard",
       "type": "string"
     },
+    "custom": {
```
carolynvs, line 35: Yikes! Thank you for adding this in, as I will be relying on it later for doing audits. 🙇‍♀️

carolynvs (approved):
LGTM

carolynvs (commented):
Wait, looking over this once more it doesn't look like cnab-go sets CNAB_LAST_REVISION at all. Are we missing an important piece of the claims implementation?

How is it supposed to be used by bundles and what should runtimes be providing? I'm suspicious that it has never been implemented and no one has noticed. Am I missing something? 😀

vdice:
> Wait, looking over this once more it doesn't look like cnab-go sets CNAB_LAST_REVISION at all. Are we missing an important piece of the claims implementation?

Thanks for checking.  Indeed, if it's not in cnab-go, we can consider it unimplemented.

> How is it supposed to be used by bundles and what should runtimes be providing? I'm suspicious that it has never been implemented and no one has noticed. Am I missing something?

I'm not sure.  It definitely smells like a relic from when these docs were first drafted and we might want to consider dropping the CNAB_LAST_REVISION requirement.

carolynvs:
@technosophos @chris-crone @jlegrone do you have any thoughts on the question of `CNAB_LAST_REVISION` having never been implemented (previous 2 comments above)? Is it safe to remove or do we need to clarify the spec around what "last" means and then implement it in cnab-go?

chris-crone:
I don't remember discussing `CNAB_LAST_REVISION` so I don't know what the use cases for it being present would be. I suggest we try to raise it on Slack and in the community meetings to see what use cases people imagine this would be useful for.

jlegrone:
I think it's ok to remove `CNAB_LAST_REVISION`. Defining what this means could be tricky; is it the revision associated with the last successful install/upgrade action? the last completed? what if there are multiple concurrent operations that all modify revision? For runtimes that need it I suggest we move it to a non-normative part of the spec for now.

Most of the use cases I can think of involve state tracking, which we can probably defer to invocation images (ie. lean on terraform state files, Helm release metadata or kubectl's last-applied annotation etc) until we have a better defined set of use cases.

The concept of a "last revision" is still important for us to define though in order for things like the "output" parameter source type to have well defined behavior. https://github.com/cnabio/cnab-spec/blob/master/810-well-known-custom-extensions.md#output

carolynvs:
I am in agreement with Jacob that defining what is the last revision to provide isn't as straightforward as it appears. My vote is to remove CNAB_LAST_REVISION and let the bundle track its own state.

vdice:
Ok, thank you all.  I will add a commit removing instances/text around CNAB_LAST_REVISION.

vdice:
Aforementioned changes added in https://github.com/cnabio/cnab-spec/pull/386/commits/3f9b8b5acc86fa03f61dc09f0aebfb111c508fe0

carolynvs (approved):
LGTM

radu-matei (approved):

carolynvs:
Seems like we have consensus on this. Are we okay to merge or do we expect different feedback by waiting until Wednesday's meeting?

vdice:
I'd be in support of merging and presenting the decisions made in this PR in Wednesday's meeting.  Even if there are requests to re-introduce/expand upon the partially-formed concept we've removed, it would need to be a separate PR.  We could then determine if it should block the 1.0.0 release or be scheduled for a future version.

chris-crone (approved):
LGTM

Issue #387, opened 2020-09-17
Cut cnab-core-1.1.0 release


vdice:
The clarification added in https://github.com/cnabio/cnab-spec/pull/384 has been deemed to be substantial enough to warrant a minor release of the core spec.  This issue tracks the process of the release.

- [x] Issue PR w/ updates relevant text bumping the cnab-core version to proposed 1.1.0 -> https://github.com/cnabio/cnab-spec/pull/388
- [x] Once merged, cut cnab-core-1.1.0-ga branch
- [x] Garner final approval from working group
- [x] Cut cnab-core-1.1.0 tag

vdice:
I've created the [cnab-core-1.1.0-ga branch](https://github.com/cnabio/cnab-spec/tree/cnab-core-1.1.0-ga) now that https://github.com/cnabio/cnab-spec/pull/388 has been approved and merged.

Question: does the approval of #388 constitute working group approval?  Or is there an additional process we want to follow before cutting the `cnab-core-1.1.0` tag?  cc @chris-crone @technosophos

PR #388, opened 2020-09-17
Updates in anticipation of cnab-core-1.1.0

vdice:
Docs and other updates in anticipation of the forthcoming/proposed cnab-core-1.1.0 release.

The main spec change prompting this minor release is the following clarification: https://github.com/cnabio/cnab-spec/issues/287

Ref https://github.com/cnabio/cnab-spec/issues/387

### DIFF
--- a/000-index.md
+++ b/000-index.md
@@ -5,7 +5,7 @@ weight: 001
 
 # CNAB Specifications
 
-1. [Cloud Native Application Bundle Core 1.0.0 (CNAB1)](100-CNAB.md)
+1. [Cloud Native Application Bundle Core 1.1.0 (CNAB1)](100-CNAB.md)
     - [The bundle.json File](101-bundle-json.md)
     - [The Invocation Image Format](102-invocation-image.md)
     - [The Bundle Runtime](103-bundle-runtime.md)
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -13,7 +13,7 @@ The `bundle.json` file is a representation of bundle metadata. It MUST be repres
 
 A `bundle.json` is broken down into the following categories of information:
 
-- The schema version of the bundle, as a string with a `v` prefix. This schema is to be referenced as `v1` or `v1.0.0`
+- The schema version of the bundle, as a string with a `v` prefix. This schema is to be referenced as `v1` or `v1.1.0`
 - The top-level package information (`name` and `version`)
   - `name`: The bundle name, including namespacing. The namespace can have one or more elements separated by a dot (e.g. `acme.tunnels.wordpress`). The left most element of the namespace is the most general moving toward more specific elements on the right.
   - `version`: Semantic version of the bundle
@@ -123,7 +123,7 @@ The following is an example of a `bundle.json` for a bundled distributed as a _t
          }
       }
    },
-   "schemaVersion":"v1.0.0",
+   "schemaVersion":"v1.1.0",
    "version":"0.1.2"
 }
 ```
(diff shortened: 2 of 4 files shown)


000-index.md
```diff
@@ -5,7 +5,7 @@ weight: 001
 
 # CNAB Specifications
 
-1. [Cloud Native Application Bundle Core 1.0.0 (CNAB1)](100-CNAB.md)
+1. [Cloud Native Application Bundle Core 1.1.0 (CNAB1)](100-CNAB.md)
```
vdice, line 8: Q: Does `CNAB1` need to change?

101-bundle-json.md
```diff
@@ -13,7 +13,7 @@ The `bundle.json` file is a representation of bundle metadata. It MUST be repres
 
 A `bundle.json` is broken down into the following categories of information:
 
-- The schema version of the bundle, as a string with a `v` prefix. This schema is to be referenced as `v1` or `v1.0.0`
+- The schema version of the bundle, as a string with a `v` prefix. This schema is to be referenced as `v1` or `v1.1.0`
```
vdice, line 16: I wasn't sure if all of these `v1.0.0` references should be updated to `v1.1.0` or not?

README.md
```diff
@@ -4,15 +4,16 @@
 
 Cloud Native Application Bundles (CNAB) are a package format specification that describes a technology for bundling, installing, and managing distributed applications, that are by design, cloud agnostic.
 
-## CNAB Core 1.0 (Final)
+## CNAB Core 1.1.0 (Final)
 
-The CNAB Working Group with the joint approval of the Executive Directors has approved the CNAB Core 1.0 specification for publication. CNAB Core 1.0 is complete.
+The CNAB Working Group with the joint approval of the Executive Directors has approved the CNAB Core 1.1.0 specification for publication. CNAB Core 1.1.0 is complete.
```
vdice, line 9: Should I have added an additional section rather than amending this one?

README.md
```diff
@@ -4,15 +4,16 @@
...
 
-For more information on the approval process, see [the process documentation](901-process.md). Further changes to CNAB Core will be considered for CNAB Core 1.1.
+For more information on the approval process, see [the process documentation](901-process.md). Further changes to CNAB Core will be considered for CNAB Core 1.2.
 
 ### Branch/Tag Structure
 
 - [master](https://github.com/cnabio/cnab-spec) is the current working draft of all specs
+- [cnab-core-1.1.0](https://github.com/cnabio/cnab-spec/tree/cnab-core-1.1.0) is the CNAB Core 1.1.0 final draft
```
vdice, line 16: This tag is yet to be created; the intention is to cut the tag after this PR is approved, merged and included in the forthcoming `cnab-core-1.1.0-ga` branch.  Ref ordering in https://github.com/cnabio/cnab-spec/issues/387

000-index.md
chris-crone: I would prefer if we have CNAB<major> mean the latest CNAB release of the <major> version (i.e.: In this case, CNAB1 means CNAB 1.1.0) as I think that usually this is what people mean when they say version <major>. If one wants to be more specific, they can do so by specifying the minor and patch versions as well.

If we choose to do this, we may need to clarify our [process versioning](https://github.com/cnabio/cnab-spec/blob/master/901-process.md#versioning).

101-bundle-json.md
chris-crone: While the schema has not changed, it is versioned with the core of the spec so I think it should match that (i.e.: 1.1.0).

README.md
chris-crone: I think amending here makes sense. We don't want to pollute the README over time with a release history.

chris-crone (approved):
LGTM

README.md
technosophos: Agree

technosophos (approved):
Sorry that it took me so long to review this, but it looks good as-is.

vdice:
@carolynvs @radu-matei @jlegrone Can we get your approval on this PR/CNAB Core 1.1.0 release?

The changelog prompting this minor release is:

- docs(101-bundle-json.md): add clarification around contentDigest value https://github.com/cnabio/cnab-spec/pull/384

PR #396, opened 2020-12-17
chore(*): pin ajv cli to known working version

vdice:
* Pins the ajv cli that we use to validate json to a known working version

It looks like the latest release(s) introduce new flags/usage that we'll need to look into when/if we bump.  (See https://github.com/cnabio/cnab-spec/issues/397)


carolynvs (approved):
PIN ALL THE THINGS! 👏

radu-matei (approved):
This is a great catch, thanks!

LGTM

PR #407, opened 2021-02-17
Add applyTo to Credentials

### LINKED ISSUE #366: Can we add applyTo to credentials too?
Currently, both parameters and outputs support the `applyTo` field which allows an author to communicate that the paramter/output is only applicable for particular actions. This is helpful especially for required parameters, so that as a user, I don't have to specify a required parameter for actions where it isn't used at all.

At first glance, being able to say "I don't need (rando cloud credentials) to perform the dry-run action" would be very user friendly. It's annoying to have to specify credentials for actions that do not use them just because install needed those credentials.

I would like to understand if `applyTo` was left off of credentials on purpose or if it was an oversight?

I can see technical / usability challenges around *CredentialSets* when each action may require a different sets of credentials depending on `applyTo`. But I am not sure I see any outright problems that would prevent us from adding `applyTo` to credentials. What do you all think?

carolynvs:
Allow restricting a credential to a list of actions, just like Parameters can do today. This allows a credential to not have to be
passed in for actions that do not use it. 

Closes #366

### DIFF
--- a/101-bundle-json.md
+++ b/101-bundle-json.md
@@ -147,7 +147,8 @@ What follows is an example of a thick bundle. Notice how the `invocationImage` a
          "path":"/etc/hostkey.txt"
       },
       "image_token":{
-         "env":"AZ_IMAGE_TOKEN"
+         "env":"AZ_IMAGE_TOKEN", 
+         "applyTo": ["install"]
       },
       "kubeconfig":{
          "path":"/home/.kube/config"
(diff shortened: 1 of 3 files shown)


vdice (approved):
LGTM.  As this is additive (and optional), we should be good with a minor bump in the bundle spec when this is merged, right?

radu-matei (approved):
LGTM

carolynvs:
Correct, this doesn't change existing behavior when applyTo is not set. We talked about it in the CNAB meeting with Matt and Chris and agreed this is a minor spec change.

PR #409, opened 2021-02-21
Support namespace and labels

### LINKED ISSUE #408: Organize data with namespaces and labels
Porter is looking to improve dependency management per this [proposal](https://github.com/getporter/porter/discussions/1441) and a key part of the implementation is that installations can be scoped by namespace and queryable by a set of labels.

I will follow up this issue with a PR to both the Claims and Credential Set specs to includes these concepts. For dependency management, I think we just need this for claim data but it's going to be useful for all CNAB tool data. For example, being able to have an installation reference a credential set that is in the same namespace.

Looking at cnab-spec's crud interfaces and types, we can generalize the "group" argument that allows us to do foreign key references and change it to a map[string]string to support labels. Since namespace needs to be understood by the storage providers in order to create unique keys, namespace should be added as an argument to any functions that accept a name.

carolynvs:
* Namespaces allow us to scope documents to a particular namespace or leave empty to define it in the global namespace.
* Labels let us define key/value pairs on data stored that we can use to query later.
* Add labels to bundles, credential sets, installations, claims, results, and outputs
* Add namespaces to credential sets, installations, claims, results, and outputs.

Bundles already had the concept of namespacing via using the format com.example.mybundle. I did not want to use this format for the other documents and opted to use a separate field to preserve existing behavior. People could have named their other documents with periods not intending for it to be interpreted as namespaces.

In the examples for Credential Sets, I change it to reference how Porter does it because Duffle is archived and won't implement the new fields.

Closes #408


101-bundle-json.md
```diff
@@ -287,6 +293,7 @@ The following fields are informational pieces of metadata designed to convey add
 
 - `description`: A short description of a bundle (OPTIONAL)
 - `keywords`: A list of keywords (OPTIONAL)
+- `labels`: Key/value pairs of type string that specify identifying attributes of the bundle.
```
vdice, line 296: Perhaps for consistency, we append `(OPTIONAL)` to this line as well.

vdice (approved):
LGTM

101-bundle-json.md
carolynvs: So you are saying `(OPTIONAL)` isn't optional, eh? 💁‍♀️

400-claims.md
```diff
@@ -105,6 +109,8 @@ The fields above are defined as follows:
 - `custom` (OPTIONAL): A section for custom extension data applicable to a given runtime.
 - `id` (REQUIRED): The claim id. A [ULID](https://github.com/ulid/spec) that MUST change with each new claim, so that every claim associated with an installation has a unique id. This is used to associate the the claim with its result(s). 
 - `installation` (REQUIRED): The name of the _installation_. This can be automatically generated, though humans may need to interact with it. It MUST be unique within the installation environment, though that constraint MUST be imposed externally. Elsewhere, this field is referenced as the _installation name_. The format of this field must follow the same format used for the `installation` field in the [bundle.json file specification](101-bundle-json.md#the-bundlejson-file).
+- `labels` (OPTIONAL): Labels are a set of key/value pairs of type string that MAY be used for querying. Labels defined on an installation MUST be included in the claim.
```
carolynvs, line 112: > Labels defined on an installation MUST be included in the claim.

This is a dirty hack because we do not have a document that represents the installation. Each tool _could_ make up their own representation but that isn't really any better. So I am having the claim persist the installation labels just like it does the installation name already.

carolynvs:
@jlegrone Do you think that label sources should be at the spec level? I'm also wondering what sources we would support initially, maybe

* bundle (labels defined by the bundle.json)
* invocation image (labels added to the installation or claim by the executing invocation image)
* user (set by the user manually)

carolynvs:
I'm going to follow Butcher's advice from https://github.com/cnabio/cnab-spec/issues/410#issuecomment-805345680 and start a new spec for this stuff.

PR #411, opened 2021-03-24
Installation State Specification

carolynvs:
After having the claims specification implemented for a while, and as new features are suggested, we've identified areas where the spec could be improved. For example, storing installation data in namespaces.

Rather than immediately bumping the CNAB Claim spec to v2, I am introducing a new specification "Installation State" that tools can implement and provide feedback on. Eventually when it stabilizes we can consider using it as the Claims v2 spec or replacing the Claims spec with this spec instead.

* Introduce installation document to store data associated only with the installation that isn't related to a particular bundle execution (claim).
* Support namespace and labels
  * Namespaces allow us to scope documents to a particular namespace or leave empty to define it in the global namespace.
  * Labels let us define key/value pairs on data stored that we can use to query later.
  * Add labels to bundles, credential sets, installations, claims, results, and outputs
  * Add namespaces to credential sets, installations, claims, results, and outputs.
  * Bundles already had the concept of namespacing via using the format com.example.mybundle. I did not want to use this format for the other documents and opted to use a separate field to preserve existing behavior. People could have named their other documents with periods not intending for it to be interpreted as namespaces.
* Expand credential sets to include parameter sets, and include namespaces and labels to the schema.


chris-crone (commented):
Thanks @carolynvs!

I still need to do a full review of this but had some time between meetings and found a couple of minor things

vdice (commented):
Looking good!

Question:  Do we want to add a `schemaVersion` field to applicable schemas and specs (installation, claim, result, cred set, param set) -- perhaps as optional?  Mentioning that tools can utilize this info for migration, compatibility, interop and other misc. purposes?

401-installation-state.md
```diff
@@ -0,0 +1,445 @@
...
+
+An installation MAY change over time, as a particular bundle is installed, then upgraded. In CNAB, each time a modifying operation is performed (Such as `install`, `upgrade`, or custom operations that are not read-only), the installation gets a new _revision_. A revision is a unique identifier that identifies the combination of an _installation_ and a modification of that installation. Revision is stored on the claim. For example, when a CNAB bundle is installed, the installation will have an initial revision ID. When that installation is upgraded, it will have a new revision ID. How revisions are used is outside of the scope of this document, but it is safe to assume that if a revision ID has changed, one or more artifacts owned by the installation has also been changed.
```
chris-crone, line 108:
```suggestion
An installation MAY change over time, as a particular bundle is installed, then upgraded. In CNAB, each time a modifying operation is performed (such as `install`, `upgrade`, or custom operations that are not read-only), the installation gets a new _revision_. A revision is a unique identifier that identifies the combination of an _installation_ and a modification of that installation. Revision is stored on the claim. For example, when a CNAB bundle is installed, the installation will have an initial revision ID. When that installation is upgraded, it will have a new revision ID. How revisions are used is outside of the scope of this document, but it is safe to assume that if a revision ID has changed, one or more artifacts owned by the installation has also been changed.
```

401-installation-state.md
```diff
@@ -0,0 +1,445 @@
...
+- `parameters` (OPTIONAL): Key/value pairs that were passed in during the operation. These are stored so that the operation can be re-run. See the [Parameters](#parameters) section below for more details.
+- `revision` (REQUIRED): The _installation_ revision. A [ULID](#ulids) that MUST change each time the installation is modified. It MUST NOT change when a [non-modifying operation](https://github.com/cnabio/cnab-spec/blob/main/101-bundle-json.md#custom-actions) is performed on the installation.
+
+#### Credentials
+
+Note that credential data is _never_ stored in a claim. For this reason, a claim is not considered _trivially repeatable_. Credentials MUST be supplied on each action. Implementations of the Installation State specification are expected to retrieve the credential requirements from the `bundle` field.
+
+#### Parameters
+
+If parameters are passed in during the operation, they MUST be stored on the claim.  The parameter data stored in a claim is _the resolved key/value pairs_ that result from the following transformation:
```
chris-crone, line 192:
```suggestion
If parameters are passed in during the operation, they MUST be stored on the claim. The parameter data stored in a claim is _the resolved key/value pairs_ that result from the following transformation:
```

401-installation-state.md
```diff
@@ -0,0 +1,445 @@
...
+  - `contentDigest` (OPTIONAL): Contains a digest, in [OCI format](https://github.com/opencontainers/image-spec/blob/master/descriptor.md#digests), which can be used to compute the integrity of the output.
+  - `generatedByBundle` (OPTIONAL): Indicates if the output was defined in the bundle and populated by the invocation image's execution. Outputs on the result can also be dynamically generated by the CNAB runtime.
+  - `labels` (OPTIONAL): [Labels](#labels) are a set of key/value pairs of type string that MAY be used for querying. Labels defined on an output in a bundle SHOULD be copied to the output's labels field on the claim result that generated the output. MUST follow the [CNAB Label Format].
+- `status` (REQUIRED): Indicates the status of the last phase transition. Valid statuses are:
+  - `cancelled`: The operation was cancelled, potentially during the operation's execution. This is an error condition.
```
chris-crone, line 252: I see that we used `cancelled` rather than `canceled` across the CNAB specification. We're consistent so I think that it's OK.

802-credential-sets.md
```diff
@@ -66,9 +86,79 @@ Similar tooling could choose to load the values by name from a database, vault,
 
 The created and modified timestamps are in [RFC 3339](https://www.ietf.org/rfc/rfc3339.txt) format.
 
-## Credential Injection
+## Parameter Mapping
+
+Parameter values can be mapped by the runtime using parameter sets, which works the same as credential sets.
```
chris-crone, line 91:
```suggestion
Parameter values can be mapped by the runtime using parameter sets, which work the same as credential sets.
```

802-credential-sets.md
```diff
@@ -66,9 +86,79 @@ Similar tooling could choose to load the values by name from a database, vault,
...
+         }
+      }
+   ]
+}
+```
+Source: [802.01-parameter-set.json](/examples/802.01-parameter-set.json)
+
+### Namespaces
+
+Credential and Parameter Sets MAY be scoped to a namespace.
```
chris-crone, line 134: Question: Should we use RFC 2119 terminology for non-normative parts of the spec? It seems we're currently doing a bit of a mix but it might make sense to use it for parts that we envisage making normative.

(Higher up, I see we use "can" and "should" which is what triggered this question)

chris-crone (commented):
Gave it a first read through and I think it's good, thanks @carolynvs!

Only minor comments and questions

401-installation-state.md
```diff
@@ -0,0 +1,445 @@
...
+---
+
+# CNAB Installation State 1.0.0
+*Draft, Mar. 2021*
+
+This specification describes CNAB Installation State. This is not part of the [CNAB Core](100-CNAB.md) specification, but is a specification describing how state related to an installation of a bundle is formatted and stored. Implementations of CNAB Core can meet the CNAB Core specification without implementing this specification. Implementations that support installation state MAY declare that they comply with CNAB Installation State 1.0.0.
+
+The Installation State specification is based on the CNAB Claims specification, extending it to cover additional data that MAY be used by a CNAB runtime. After we have implemented and vetted this specification, one possible outcome would be to replace the Claims specification with this one.
+
+A _host environment_ is an environment, possibly shared between multiple CNAB tools, that provides persistence for CNAB documents. For example, a filesystem may contain a record of claims that two different CNAB clients share. Or a database may contain the environment that is shared by multiple tools at different locations in the network. Each of these is a CNAB host environment.
```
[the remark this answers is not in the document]
carolynvs: This has been the term since the beginning of the CNAB spec (it's taken from the claims spec). I also think it is unclear and am happy to come up with a different name for it, as long as we understand that changing the name could lead to two different specs with different terms for the same thing.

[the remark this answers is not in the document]
carolynvs: Also the host environment can contain more than just installation state data. It also has credential/parameter stores, and potentially other documents used to the cnab tool that aren't covered in the spec.

401-installation-state.md
carolynvs: 🤦‍♀️ I just checked in cnab-go and we are storing `canceled` on the claims. Let me start another issue to track this goofup.

carolynvs: https://github.com/cnabio/cnab-spec/issues/412

802-credential-sets.md
carolynvs: I think it would be easier to maintain if we used the same language throughout the spec. Just so that as text is copied about you don't have to change it, and also so it's easier to spot when we mess up.

What do you think @technosophos?

carolynvs:
I'm so glad you remembered about the schema version! I'll make sure that all the docs have that going forward.

401-installation-state.md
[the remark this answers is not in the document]
chris-crone: I'd forgotten that we used it elsewhere. You're right, we should be consistent and if it isn't clear we can discuss it as a separate issue.

802-credential-sets.md
chris-crone: +1

carolynvs:
@vdice I have added schemaVersion to the following documents:

* installation
* claim
* claim result
* parameter set
* credential set

For the installation/claim/result docs, the value will be the version # of the installation state spec. For parameter/credential sets, it's a bit fuzzier since neither are actual formal specs with a version. Right now we have been tagging the spec following the cnab core version number, https://github.com/cnabio/cnab-spec/releases/tag/cnab-credentialsets-1.0.0-DRAFT%2Bb6c701f. Now that it is going to encompass both parameter and credential sets, I think we need to change this.

What if I made a new formal spec based on parameter and credential sets and name it ValueSet Specification (that's the name from cnab-go)? Then we can version it going forward. Thoughts?

vdice:
@carolynvs Thanks for adding!  Looks great.

> What if I made a new formal spec based on parameter and credential sets and name it ValueSet Specification (that's the name from cnab-go)? Then we can version it going forward. Thoughts?

Definitely in support.  Are we thinking of adding to this same PR?

carolynvs:
I will take the changes I made to the non-normative cred spec out of this PR and move it into a separate PR that makes a real spec and adds schemaVersion, namespace and labels.

vdice (commented):
Do we want to add another section in `scripts/validate.sh` to validate the `401.01-installation.json` example against the associated spec?

vdice (approved):
LGTM!

carolynvs:
@radu-matei @jlegrone Can you take a look? I have incorporated Chris and Vaughn's feedback. The changes to parameter/credential sets have been moved into a new PR (incoming) but I still need a second review from a code owner.

carolynvs:
After discussing with @technosophos and @vdice, I have realized that I was putting too many client concerns into this spec. I was losing sight that we only need to specify **runtime** concerns, or anything necessary that a bundle can be executed by any conformant runtime. Clients can extend the documents here and it doesn't need to be standardized.

We've always shared some common CRUD functionality in cnab-go, and used by the action package, which helped drive the confusion for me on what belongs or not. It made sense to share common structs for what a claim should look like, but things like the CRUD data access layer shouldn't have been in there IMO. It forced a tighter level of agreement between consumers of that library than was necessary.

I'm going to do another pass at this to just define the documents and terms necessary for what we need at runtime, which AFAIK is the concept of defining an installation within an namespace with associated labels that can be used to filter them. This is necessary for our plans for the dependencies spec, where a bundle can take a dependency on an existing installation that matches certain criteria. For example: "I need a mysql database in my namespace, with the following matching labels", and the runtime can use that criteria to decide to either create a new installation or reuse an existing one.

carolynvs:
I'm going to close this and submit a new PR with the changes when they are ready.

PR #414, opened 2021-05-17
Allow numbers again

### LINKED ISSUE #413: revisit canonical json and non-integer numbers
I am confused about the official spec for canonical json and whether or not it supports non-integer numbers. I think [this is the spec ](https://tools.ietf.org/html/rfc8785) and they do say that big/high precision numbers should be stored as strings and parsed in the consumer. They *do not* rule out all non-integer numbers though.

That spec also says that this go implementation is conformant: https://github.com/cyberphone/json-canonicalization/tree/master/go/src/webpki.org/jsoncanonicalizer

We are currently using docker's canonical json library which does not support non-integer numbers. It has a [hard check](https://github.com/docker/go/blob/d30aec9fd63c35133f8f79c3412ad91a3b08be06/canonical/json/encode.go#L524) when marshaling a float that it is a whole number.

I have tried the other library and it round trips the value "1.5" (which is what I have users asking for) fine.

When [we changed the spec](https://github.com/cnabio/cnab-spec/issues/256) to exclude non-integer numbers that decision was based on a old?bad? copy of the canonical json spec (I was looking at the spec referenced by the docker library http://wiki.laptop.org/go/Canonical_JSON) instead of the one above. I now believe that was a mistake and that we should revert that change and update cnab-go to use a different library.

Tagging @chris-crone and @technosophos to weigh in.

carolynvs:
Revert the changes in #257 which updated the spec to disallow non-integer values. We misunderstood the canonical json spec and were using a library that didn't support numbers.

I have an open PRs for cnab-go which tested out switching to a library that support numbers in canonical json and it works great.

* https://github.com/cnabio/cnab-go/pull/247
* https://github.com/cnabio/cnab-go/pull/248

So now fields such as default, maximum, minimum, etc can use numbers again! 🎉

Closes #413


carolynvs:
@technosophos @chris-crone I believe this would be a patch to the spec version? I think we will want to bump the spec version after this is merged.

101-bundle-json.md
```diff
@@ -415,7 +415,7 @@ of how to use `definitions` along with `parameters` and `outputs` can be seen in
     - `required`: Parameter validation requiring the properties named in the user-provided object include the specified list of properties. MUST be an array of strings. (OPTIONAL)
     - `then`: Parameter validation requiring that the user-provided value match the specified schema. Only matches if the user-provided value matches the schema provided in the `if` property. MUST be a JSON schema. (OPTIONAL)
     - `title`: Short, human-readable descriptive name for the field. Can be used to decorate a user interface. MUST be a string. (OPTIONAL)
-    - `type`: Parameter validation requiring that the user-provided value is either a "null", "boolean", "object", "array", "string", or "integer". MUST be a string or an array of strings with unique elements. If you need to represent another numeric type, upscale to an integer or use a string type and convert within your bundle. (OPTIONAL)
+    - `type`: Parameter validation requiring that the user-provided value is either a "null", "boolean", "object", "array", "string", or "number". MUST be a string or an array of strings with unique elements. (OPTIONAL)
```
vdice, line 418: We can keep "integer" as well, right?  I think it's a valid JSON Schema construct distinct from "number".

vdice (commented):
Oooh, nice to be able to get rid of the custom definitions schema 🎉 

Only comment is around continuing support for `integer` in addition to `number`.

vdice (approved):

radu-matei (approved):
LGTM, with the mention that the canonical JSON library used in Signy will have to be updated as well.

carolynvs:
I'll submit a PR to signy as I already went through the switch for cnab-go.

Here are the follow-up PRs to this spec change for cnab-go:
https://github.com/cnabio/cnab-go/pull/247
https://github.com/cnabio/cnab-go/pull/248

PR #416, opened 2021-05-28
Add back definitions.schema.json

carolynvs:
I didn't realize that the URLs embedded in our schema are redirects to the **latest** code on main. I should not have removed
definitions.schema.json from the repository when we no longer used it.


technosophos (approved):
This is purely restoring a file that was here before and was deleted during a larger PR. Since the missing file is breaking real users on a Friday (and since there are no changes between this and the previously deleted version), I'm good with merging right away without a second LGTM.

PR #420, opened 2021-09-14
Use status canceled instead of cancelled

### LINKED ISSUE #412: Inconsistent spelling of canceled state
In the spec we consistently use the spelling `cancelled`. However in cnab-go we have been using the alternative spelling `canceled` when we store state on the claim.

We should pick a spelling (british vs american) and use the same one everywhere. I lean towards sticking with `canceled` since that's what is stored in user's data right now, and update the spec to fix the misspelling.

carolynvs:
The spec originally used a variation of the word "canceled" and it was spelled "cancelled". However all of the CNAB implementations (duffle, porter, Data Dog) all used "canceled".

We have agreed to use the term that is already in everyone's data, and this updates the spec to use the term "canceled".

Fixes #412


technosophos (approved):
This is an odd one. Spelling changes are considered errata, and do not require a formal spec revision. However, changing a field name does. However (again), in this case the spelling in the spec was changed to match the spelling in all known CNAB implementations.

My feeling is that this should be considered an erratum, and should not require a formal spec revision.

vdice (approved):

jlegrone (approved):

